PR #10, opened 2019-07-10
feat(dataset-view): layout commit list items

chriswhong:
Basic layout for items the history tab

Qri_Desktop


app/components/DatasetSidebar.tsx
```diff
@@ -12,6 +12,47 @@ const FileRow: React.SFC<FileRowProps> = (props) => {
   )
 }
 
+interface HistoryListItemProps {
+  commitTitle: string
+  avatarUrl: string
+  userTimeMessage: string
+}
+
+const HistoryListItem: React.SFC<HistoryListItemProps> = (props) => {
```
ramfox: Should be `React.FunctionComponent`

app/scss/_dataset-container.scss
```diff
@@ -133,6 +133,7 @@
     width: 100%;
```
ramfox: Didn't we change this to "_dataset.scss" and move the previous "_dataset.scss" to "_dataset-old.scss"?

ramfox: What do you think about moving the `.file-row` and `.history-list-items` stuff into it's own scss file. I think I'm liking the "one file one component" structure. Thoughts?

chriswhong: I re-did the list-item stuff just now on another branch so it's cleaner and more centralized.  I think it's fine in `dataset.scss` as it probably won't be used outside of this view, but is shared by more than one child component

PR #11, opened 2019-07-11
Logo

chriswhong:
Adds a 32x32 qri logo to the top left of the dataset view
Qri_Desktop_and_Comparing_master___logo_·_qri-io_desktop


app/containers/DatasetContainer.tsx
```diff
@@ -3,6 +3,7 @@ import { Resizable } from '../components/resizable'
 import DatasetSidebar from '../components/DatasetSidebar'
 
 const defaultSidebarWidth = 250
+const logo = require('../assets/qri-blob-logo-tiny.png') //eslint-disable-line
 
 export default class DatasetContainer extends React.Component<{}, { showDatasetList: boolean, sidebarWidth: number }> {
```
ramfox: Didn't we change this to `Dataset` and move it to `/components/Dataset`?

app/scss/_dataset-container.scss
```diff
@@ -22,7 +22,7 @@
     display: inline-block;
```
ramfox: Mentioned this in the other pr, but this file should be `_dataset.scss`, and the old `_dataset.scss` should be `_dataset-old.scss`

package.json
```diff
@@ -137,7 +137,7 @@
     "electron-devtools-installer": "2.2.4",
     "enzyme": "3.10.0",
     "enzyme-adapter-react-16": "1.14.0",
-    "eslint": "6.0.1",
+    "eslint": "^6.0.1",
```
ramfox: As a convention, we put exact versions, should go back to ` "eslint": "6.0.1",`

package.json
```diff
@@ -137,7 +137,7 @@
...
     "eslint-config-standard-with-typescript": "7.1.0",
     "eslint-loader": "^2.2.1",
```
ramfox: Same comment as above: should be changed to `"eslint-loader": "2.2.1",`

app/containers/DatasetContainer.tsx
chriswhong: I'll rename it again now.

package.json
chriswhong: fixed

package.json
chriswhong: fixed

PR #12, opened 2019-07-11
Dataset list

chriswhong:
Adds a dataset list pane with simple text-based filter using dummy data.

Cleans up scss a bit for consistent look and feel and common code across sidebar lists.

Jul-11-2019 12-42-48


ramfox:
Wait one thought! What if we changed the 'history/status' bottom bar color from that blue to our dark blue?

PR #14, opened 2019-07-12
Redux setup

chriswhong:
- Adds sets up redux actions, reducers, state, and containers
- Refactors `Dataset` and `DatasetSidebar` to use containers, removes the stubbed-in local state


chriswhong:
Closing, will go a different way

PR #20, opened 2019-07-15
feat: add first-pass model for state tree

chriswhong:
Adds a first-pass of the app's redux state tree.


package.json
```diff
@@ -114,6 +114,7 @@
     "@types/enzyme": "3.10.2",
     "@types/history": "4.7.2",
     "@types/jest": "24.0.15",
+    "@types/json-schema": "^7.0.3",
```
ramfox: should be exact version, remove the caret

PR #21, opened 2019-07-17
feat(api): initial pass at API middleware

b5:
working on fitting api middleware to our refactored state tree layout.

I'm also realizing that I renamed your component from App to `Onboarding` in the process @ramfox, holler if that's an issue.

### DIFF
--- a/app/containers/MyDatasetsContainer.tsx
+++ b/app/containers/MyDatasetsContainer.tsx
@@ -1,0 +1,15 @@
+import { connect } from 'react-redux'
+import { fetchMyDatasets } from '../actions/api'
+import { DatasetList } from '../components/DatasetList'
+
+const mapStateToProps = (state: any) => {
+  return {
+    value: state.myDatasets
+  }
+}
+
+const actions = {
+  fetchMyDatasets
+}
+
+export default connect(mapStateToProps, actions)(DatasetList)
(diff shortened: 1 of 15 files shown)


app/containers/MyDatasetsContainer.tsx
```diff
@@ -0,0 +1,15 @@
...
+  return {
+    value: state.myDatasets
+  }
+}
+
+const actions = {
+  fetchMyDatasets
+}
+
+export default connect(mapStateToProps, actions)(DatasetList)
```
chriswhong: Did you mean to leave this in?  Should the container name match the component? (DatasetList !== MyDatasetsContainer)  You're also passing in `fetchMyDatasets` but not calling it from the component.

b5: it's possible to name this at the import site for devs:
```js
import DatasetListContainer from '../containers/DatasetListContaner'
```

but you're right the name of the component itself in React won't be `DatasetListContainer`. I'll fix.

As for not calling `fetchMyDatasets`, I was assuming the next step would be to pull the demo-only calls out of `OnboardingContainer` and into a component lifecycle hook in `DatasetList`. I'll make a note of this.

chriswhong: NM, I see how you used the container, I'll work this out later

PR #26, opened 2019-07-23
Component content

chriswhong:
- Adds HandsOnTable
- Adds Commit Detail Component
- Adds localstorage to persist selections


app/actions/api.ts
```diff
@@ -89,11 +89,39 @@ export function fetchWorkingDataset (): ApiActionThunk {
   }
 }
 
+export function fetchCommitDetail (): ApiActionThunk {
+  return async (dispatch, getState) => {
+    const { selections } = getState()
+    const { commit } = selections
```
ramfox: Do we know for sure that every time we pull `selections` out of state, it will be non-null?

If we don't know that for sure, then line 95 might panic. 

Unless having commit and selections be Typed means that we won't get null values here, which would be awesome.

This is usually what selectors are for, obscuring away all the ugly but safe checking that something exists before destructuring it.

app/actions/api.ts
```diff
@@ -103,7 +131,16 @@ export function fetchWorkingHistory (): ApiActionThunk {
           peername: selections.peername || 'me',
           name: selections.name || 'world_bank_population'
         },
```
ramfox: Is this a default for right now, or is this because we are going to ship qri with `me/world_bank_population`?

app/components/Body.tsx
```diff
@@ -0,0 +1,44 @@
...
+    this.state = {}
+  }
+
+  static getDerivedStateFromProps (nextProps: BodyProps) {
+    const { workingDataset } = nextProps
+    const { value, bodyLoading: isLoading } = workingDataset
+    const { body } = value
+    if (
+      body === undefined &&
+      isLoading === false
```
ramfox: If body is empty, will it come back as [] or {}?

Also, if body is undefined and isLoading is false because of an error, and the error is never handled or reacted to, this will become an infinite loop.

app/components/Dataset.tsx
```diff
@@ -119,7 +136,7 @@ export default class Dataset extends React.Component<DatasetProps> {
           <div
             className={'current-dataset header-column ' + expandedClass}
```
ramfox: wanted to draw your attention to a package I've been using called `classnames`

```
import * as classNames from 'classnames'

<div
  className={classNames('current-dataset', 'header-column',  { 'expanded': showDatasetList } )}
/>
```

`classNames` would return 'current-dataset header-column expanded' if showDatasetList is true, and 'current-dataset header-column' if false.

then you could remove line 101.

especially useful when there is a ton of conditional formatting.

app/components/DatasetSidebar.tsx
```diff
@@ -1,36 +1,43 @@
...
       statusColor = 'transparent'
   }
+
+  const selected = props.selected ? 'selected' : ''
+  const disabled = props.disabled ? 'disabled' : ''
   return (
     <div
-      className={`sidebar-list-item sidebar-list-item-text ${props.selected && 'selected'}`}
-      onClick={() => { props.onClick('component', props.name) }}
+      className={`sidebar-list-item sidebar-list-item-text ${selected} ${disabled}`}
```
ramfox: classNames would be super useful here:

`classNames('sidebar-list-item', 'sidebar-list-item-text',  { 'selected': props.selected, 'disabled': props.disabled })`

app/scss/style.scss
```diff
@@ -30,3 +30,5 @@
 @import "dataset";
 @import "welcome";
 @import "flex";
+
+@import '~handsontable/dist/handsontable.full.css';
```
ramfox: dope, interesting

package.json
```diff
@@ -175,8 +175,10 @@
     "webpack-merge": "4.2.1"
   },
   "dependencies": {
+    "@handsontable/react": "3.0.0",
     "electron-debug": "3.0.1",
     "font-awesome": "4.7.0",
+    "handsontable": "^7.1.0",
```
ramfox: this is probably going to change anyway 🤣, but the caret should be removed

app/actions/api.ts
chriswhong: I'll take it out, it was a placeholder.

app/actions/api.ts
chriswhong: We've got logic in place to set `selections.commit` before this action is fired.  I think typescript will just enforce a string, which is empty by default.

app/components/Dataset.tsx
chriswhong: added!

app/components/DatasetSidebar.tsx
chriswhong: added!

package.json
chriswhong: Downgraded to v6.2.2

app/components/Body.tsx
chriswhong: I think we should punt on this for now, and take a closer look at error states and notifications for all of our api calls together soon.  Opened an issue for this: #27

PR #30, opened 2019-07-24
Cw various improvements

chriswhong:
Various small features and bugfixes

fix: default widths of dataset list and header should match sidebar
feat: highlight current dataset in list
fix: resize qri icon
feat: consume body status from api
fix: increase maincontent overlay z-index
feat: add button to open linked directory
feat: add headers to body table


app/components/Dataset.tsx
```diff
@@ -128,16 +129,40 @@ export default class Dataset extends React.Component<DatasetProps> {
...
+    )
+
     return (
       <div id='dataset-container'>
         <div className='header'>
           <div
             className={classNames('current-dataset', 'header-column', { 'expanded': showDatasetList })}
             onClick={toggleDatasetList}
-            style={{ width: datasetSidebarWidth }}
+            style={{ width: `${datasetSidebarWidth}px` }}
```
ramfox: did this error? you should be able to just use `datasetSidebarWidth` if it's a number.

app/components/Dataset.tsx
```diff
@@ -177,7 +203,7 @@ export default class Dataset extends React.Component<DatasetProps> {
           showDatasetList && (
             <div
               className='dataset-list'
-              style={{ width: datasetSidebarWidth }}
+              style={{ width: `${datasetSidebarWidth}px` }}
```
ramfox: same q as above

app/components/Dataset.tsx
chriswhong: Good catch, they did error but only on the first render after refreshing the app.  I am realizing that it is because they are being pulled out of localstorage as strings.  Will fix to convert the initialstate to a number.

PR #70, opened 2019-07-29
fix: set default working dataset if none selected

### LINKED ISSUE #45: [1] bug: always have a default for working dataset in selection
After we load myDatasets & if there is no `name` in `state.selection`, choose the first dataset in our `myDatasets` list as default. Leave empty if `myDatasets.value.length === 0`

chriswhong:
- Moves `DatasetContainer` to App, where it is conditionally rendered if `hasDatasets` is true.  This means we aren't rendering it until after the call to `/list` is successful and there are datasets.
- Updates `selections` reducer to set the first dataset returned from `/list` as the selected dataset if there is none already selected.
- Removes `Router` from `index.tsx`, as we are not using routing at all for now (routing was removed in 77782e92f181cc194b5847fbb7cb63fd364fd66b).  `app/routes.tsx` remains in place for whenever we turn routing back on.

Closes #45

### DIFF
--- a/app/reducers/selections.ts
+++ b/app/reducers/selections.ts
@@ -1,19 +1,22 @@
 import { AnyAction } from 'redux'
 import { Selections } from '../models/store'
 import store from '../utils/localStore'
+import { apiActionTypes } from '../store/api'
 
 export const SELECTIONS_SET_ACTIVE_TAB = 'SELECTIONS_SET_ACTIVE_TAB'
 export const SELECTIONS_SET_SELECTED_LISTITEM = 'SELECTIONS_SET_SELECTED_LISTITEM'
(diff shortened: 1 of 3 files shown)


app/reducers/selections.ts
```diff
@@ -44,6 +47,18 @@ export default (state = initialState, action: AnyAction) => {
       store().setItem('name', name)
       return Object.assign({}, state, { peername, name })
 
+    case LIST_SUCC:
+      // if there is no peername + name in selections, use the first one on the list
+      if (state.peername === '' && state.name === '') {
+        const { peername: firstPeername, name: firstName } = action.payload.data[0]
+        console.log(action.payload.data)
```
ramfox: stray console.log

chriswhong: fixed

PR #80, opened 2019-07-30
40 handle commit

### LINKED ISSUE #40: [5] Handle Commit
When a user commits, the button currently triggers the appropriate action and API call, but the UI does not change.

On commit, the UI should refresh status & history for the working dataset

### LINKED ISSUE #76: [3] Create Toast component to show users messages and error messages
* [ ] create Toast component, with alternate styling for error and message
* [ ] create setMessage(message: string) call that has a timeout to reset after a few seconds
* [ ] create setErrorMessage(error: string) call with no timeout

chriswhong:
- Adds toast component (Closes #76)

Toast is stored in the state tree as `ui.toast`

`openToast(type, message)` opens the toast.  Valid types are 'success' and 'error'
`closeToast()` closes the toast

The `Toast` component calls `closeToast()` after 3 seconds (adjustable via props)

- Handles save button (Closes #40)

Clicking save sets `mutations.save.isLoading` to true, shows a spinner in place of the submit button.  

The `ui` reducer is listening for success or failure of save, and opens the toast with a message.

The `Dataset` component is also listening for save, and will fetch new `workingDataset` data on success.


app/reducers/mutations.ts
```diff
@@ -0,0 +1,72 @@
...
+    isLoading: false,
+    error: null
+  }
+}
+
+export const MUTATIONS_SET_SAVE_VALUE = 'MUTATIONS_SET_SAVE_VALUE'
+
+export const [SAVE_REQ, SAVE_SUCC, SAVE_FAIL] = apiActionTypes('save')
+
+const mutationsReducer: Reducer = (state = initialState, action: AnyAction): Mutations => {
```
ramfox: Since this works I think we should leave it in, but we can definitely refactor to generalize this reducer and any mutation actions.

ramfox: Since this works I think we should leave it in, but we can definitely refactor to generalize this reducer and any mutation actions.

app/models/store.ts
```diff
@@ -37,6 +57,7 @@ export interface UI {
   showDiff: boolean
   datasetSidebarWidth: number
   commitSidebarWidth: number
+  toast: Toast
```
ramfox: if we are adding `toast` to the `ui` interface, then we don't need `message` or `errorMessage`

app/reducers/ui.ts
```diff
@@ -56,6 +66,46 @@ export default (state = initialState, action: AnyAction) => {
...
+      return {
+        ...state,
+        toast: {
+          type: toastType,
+          message,
+          visible: true
+        }
+      }
+
+    case UI_CLOSE_TOAST:
```
ramfox: Closing the `Toast` should also clear the message and type. 

Any thought about removing `visible`, and just using whether or not `message` has content to determine whether the toast is visible or not?

app/reducers/mutations.ts
chriswhong: I'd love to hear what you propose, it's come up several times that it's not a 1-1 relationship between actions and reducers.

app/models/store.ts
chriswhong: fixed

app/reducers/ui.ts
chriswhong: There is an animation happening when visible changes to false.  If we reset message and type when the action happens, the changes are immediately visible as the toast is animating out... I don't think it matters to leave these set, they just need to be changed on opening the next toast.

PR #82, opened 2019-07-31
43 body state

### LINKED ISSUE #43: [3] Move Body higher up in the state tree
`body` currently exists in both `workingDataset.value` and `commitDetails.value`, both bodies are never visible at the same time, and it will make things simpler to move body up to a top-level state tree key.  Whether a dataset body or a historic (commit) body, the body component should fetch the data it needs and store it separately from the rest of the current dataset or commit context.

```
{
  body: {
    pages: ...
    value: [...]
  }
}
```

This issue is complete when body has been moved to its new location in the state tree, and all body-related containers, components, actions, and reducers are working as expected.

chriswhong:
Refactors state for `workingDataset` and `commitDetails` (which are identical except for `history`).  Both now have a top-level key `components` with keys `meta`, `body`, and `schema`, in which we store data needed to render the respective components in the app.

We no longer have the issue we faced where the results from two separate API calls were intermingling when we updated state.  

The issue (#43) said to move body up to its own top-level key in the state, using the same state for both dataset body and commit body.  This implementation stores the two bodies separately, eliminating the need for an API call if the user is simply switching between the status and history tabs.  

This PR also ensures we are only calling for new data when we need it, which will make for a nicer UX because we aren't showing unnecessary loading states.  Things like `body` can persist in state until the user actually switches to a new dataset or a new commit on the working dataset.

Closes #43


app/reducers/commitDetail.ts
```diff
@@ -7,24 +7,62 @@ const initialState: CommitDetails = {
...
-      return Object.assign({}, state, { name, path, peername, published, value })
+      const { name, path, peername, published, dataset } = action.payload.data
+      return {
+        ...state,
+        name,
+        path,
+        peername,
+        published,
+        isLoading: false,
+        components: {
```
ramfox: Do you see these as reflecting components as we display them visually, or the components in the dataset?

For example, let's say we decide we need to show details on the frontend based on the `structure` component, is this section where we store the structure component, or is this really supposed to reflect the components we show on the app?

app/actions/api.ts
```diff
@@ -243,13 +242,22 @@ export function fetchWorkingStatus (): ApiActionThunk {
   }
 }
 
-export function fetchBody (): ApiActionThunk {
+export function fetchBody (history: boolean): ApiActionThunk {
```
ramfox: This should be split into two actions `fetchBody` and `fetchHistoryBody`

Then, in the `BodyContainer`, you can switch on the `history` props, and send the `fetchBody` action if it's the dataset body, or `fetchHistoryBody` if it's the dataset history body.

app/components/Body.tsx
```diff
@@ -1,12 +1,13 @@
...
-import { WorkingDataset } from '../models/store'
 
 interface BodyProps {
-  workingDataset: WorkingDataset
+  value: any[]
+  isLoading: boolean
   headers: any[]
-  fetchBody: () => Promise<ApiAction>
+  history: boolean
+  fetchBody: (history: boolean) => Promise<ApiAction>
```
ramfox: Making the above change, we would remove `history`, `fetchBody` would have the signature `() => Promise<ApiAction>`

app/reducers/commitDetail.ts
chriswhong: I think these are where we put data that feeds a "main content" area in the app.  Meta, Body, and Schema main content react components, should have all they need by looking at their respective keys in workingDataset.components

app/components/Body.tsx
chriswhong: Gotcha, will change!

app/actions/api.ts
chriswhong: Will do!

PR #83, opened 2019-08-01
feat(backend): Run backend qri at startup, if present in backend/

dustmop:


ramfox:
wait hold on, not working on my machine!

dustmop:
In order to test on your machine, you need to manually copy the `qri` binary into `desktop/backend/` first. Did you try that? Let me know if it's still not working.

ramfox:
> In order to test on your machine, you need to manually copy the qri binary into desktop/backend/ first. Did you try that? Let me know if it's still not working.

That fixed it! dope

PR #94, opened 2019-08-05
refactor(errors, loading): fix the flow of the app

### LINKED ISSUE #27: [8] API Call Loading & Error States + UI
Loading States
- Add spinners where applicable to indicate that the app is fetching data

Error States
- Update the state tree on error
- Add UI for notifying the user when something has gone wrong

### LINKED ISSUE #46: [2] bug: addDataset and createDataset should change the working dataset
after we addDataset or createDataset, we should change the selectors and fetch what we need for the working dataset.

### LINKED ISSUE #85: [1] `pingApi` action should not set apiConnection on failure
only `setApiConnection` action should

### LINKED ISSUE #91: Create toast component
params:
in: boolean
type: message | success | failure

just a div w/ toast styling wrapped by a `CSSTransition`

### LINKED ISSUE #101: Handle when dataset has no history & when dataset is not FSI linked
When a dataset is not linked to the FSI, then the concept of 'status' does not mean anything.
When a dataset is linked to the FSI, but no commits have been made, 'history' does not mean anything.

Based on the API change to the `/history` endpoint that returns a 422 error (or a 500 error with `repo: no history`) as the error, when the dataset has no history, and since we already know when a dataset is linked or not, we can make these adjustments!

We pass down hasHistory and linkPath to dataset. If hasHistory === false, and the activeTab is 'history', we should set the active tab to 'status', if linkPath === '', and the activeTab is 'status', we should set the active tab to 'history.

We pass down hasHistory and linkPath to datasetSidebar. If hasHistory === false, don't show the history tab. If linkPath === '', don't show the status tab.

### LINKED ISSUE #102: we need to store `isLinked` in localStorage for the dataset in session.
If we have a dataset selected, and it isn't in our list of datasets yet (because of pagination), we don't know if we need to indicate `fsi=true` on our endpoints.

Rather than match the dataset to one in our `myDatasets` list, we should keep an indication of whether this dataset is linked in the selections. Then, any time we need to send an api request, we can determine if it is linked using the selection.

ramfox:
closes #85 
closes #91 

closes #46 create dataset and add dataset should send you to that dataset
closes #27 handling errors and loading status
closes #101 handle when dataset has no history or dataset is not linked 
closes #102 store isLinked in selections & localStorage 

This pr generally fixes issues flipping between commits and datasets! Makes sure a dataset has certain parts before allowing the app to display those parts. Adds sensible alternatives for default components to show.

Best way to play with this is to have some datasets that aren't linked, a dataset that is linked and has history, and a dataset that is linked but has no history, some datasets that have all the components, others that don't. We shouldn't be erroring when something is missing anymore.

This pr depends on a specific branch: [no-history](https://github.com/qri-io/qri/tree/no-history)
More specifically, it relies on [no-history-kasey](https://github.com/qri-io/qri/tree/no-history-kasey), which adds two commits to dustin's branch to smooth out some errors.

### DIFF
--- a/app/actions/api.ts
+++ b/app/actions/api.ts
@@ -44,7 +44,7 @@ export function fetchWorkingDatasetDetails (): ApiActionThunk {
     const { history } = workingDataset
     const { commit } = selections
 
-    if (commit === '' || !history.value.some(c => c.path === commit)) {
+    if (history.value.length !== 0 && (commit === '' || !history.value.some(c => c.path === commit))) {
       await dispatch(setSelectedListItem('commit', history.value[0].path))
     }
     return response
@@ -88,23 +88,18 @@ export function fetchMyDatasets (page: number = 1, pageSize: number = pageSizeDe
 
 export function fetchWorkingDataset (): ApiActionThunk {
   return async (dispatch, getState) => {
-    const { selections, myDatasets } = getState()
-    const { peername, name } = selections
-
-    // find the selected dataset in myDatasets to determine isLinked
-    const match = myDatasets.value.find((d) => d.name === name && d.peername === peername)
-
-    const params = (match && match.isLinked) ? { fsi: true } : {}
+    const { selections } = getState()
+    const { peername, name, isLinked } = selections
 
     const action = {
       type: 'dataset',
       [CALL_API]: {
         endpoint: 'dataset',
         method: 'GET',
-        params,
+        params: { fsi: isLinked },
         segments: {
-          peername: selections.peername,
-          name: selections.name
+          peername,
+          name
         },
         map: (data: Record<string, string>): Dataset => {
           return data as Dataset
@@ -198,14 +193,16 @@ export function fetchWorkingHistory (page: number = 1, pageSize: number = pageSi
       return new Promise(resolve => resolve(NO_ACTION))
     }
     const { selections } = getState()
+    const { peername, name, isLinked } = selections
     const action = {
       type: 'history',
       [CALL_API]: {
         endpoint: 'history',
         method: 'GET',
+        params: { fsi: isLinked },
         segments: {
-          peername: selections.peername,
-          name: selections.name
+          peername: peername,
+          name: name
         },
         pageInfo: {
           page,
@@ -231,14 +228,16 @@ export function fetchWorkingHistory (page: number = 1, pageSize: number = pageSi
 export function fetchWorkingStatus (): ApiActionThunk {
   return async (dispatch, getState) => {
     const { selections } = getState()
+    const { peername, name, isLinked } = selections
     const action = {
       type: 'status',
       [CALL_API]: {
         endpoint: 'status',
         method: 'GET',
+        params: { fsi: isLinked },
         segments: {
-          peername: selections.peername,
-          name: selections.name
+          peername: peername,
+          name: name
         },
         map: (data: Array<Record<string, string>>): ComponentStatus[] => {
           return data.map((d) => {
@@ -259,7 +258,7 @@ export function fetchWorkingStatus (): ApiActionThunk {
 export function fetchBody (page: number = 1, pageSize: number = bodyPageSizeDefault): ApiActionThunk {
   return async (dispatch, getState) => {
     const { workingDataset, selections } = getState()
-    const { peername, name } = selections
+    const { peername, name, isLinked } = selections
     const { path } = workingDataset
 
     if (workingDataset.components.body.pageInfo.fetchedAll) {
@@ -275,6 +274,7 @@ export function fetchBody (page: number = 1, pageSize: number = bodyPageSizeDefa
           page,
           pageSize
         },
+        params: { fsi: isLinked },
         segments: {
           peername,
           name,
--- a/app/actions/api.ts
+++ b/app/actions/api.ts
@@ -7,6 +7,10 @@ import { setSelectedListItem } from './selections'
 
 const pageSizeDefault = 15
 
+// use NO_ACTION when you need to skip/debounce unneccessary calls to the api
+// it will still register as a success
+export const NO_ACTION: Action = {type: 'NO_ACTION_SUCCESS'}
+
 export function pingApi (): ApiActionThunk {
   return async (dispatch) => {
     const pingAction: ApiAction = {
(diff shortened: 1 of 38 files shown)


app/components/App.tsx
```diff
@@ -29,6 +29,7 @@ interface AppProps {
   fetchSession: () => Promise<ApiAction>
   fetchMyDatasets: (page?: number, pageSize?: number) => Promise<ApiAction>
   addDataset: (peername: string, name: string) => Promise<ApiAction>
+  setWorkingDataset: (peername: string, name: string, isLinked: boolean) => Promise<ApiAction>
```
chriswhong: Someday we need to define types for these actions once and export them, we have been doing this a lot.  No issue with this PR, just making a note of it.

app/components/App.tsx
```diff
@@ -61,16 +62,16 @@ export default class App extends React.Component<AppProps, AppState> {
...
         iter++
-      }, 10000)
+        if (this.props.apiConnection === 1) clearInterval(backendLoadedCheck)
+        if (iter > 20) {
+          this.props.setApiConnection(-1)
+          clearInterval(backendLoadedCheck)
+        }
+      }, 750)
     }
-    this.props.pingApi()
```
chriswhong: Sorry, I left this in when I had set the interval higher for debugging.

app/components/ChoosePeername.tsx
```diff
@@ -32,13 +36,22 @@ const ChoosePeername: React.FunctionComponent<ChoosePeernameProps> = (props: Cho
...
     })
-      .then(async () => onSave(newPeername))
-      .catch(() => {
-        setLoading(false)
-        setError('some error occured')
+      .then(async () => {
+        setPeername(newPeername)
+          .then((action) => {
+            setLoading(false)
+            // TODO (ramfox): possibly these should move to the reducer
```
b5: I think your instinct to move this somewhere else is right. This feels more like a it should be a compound action than reducer logic. Let's save it for a subsequent PR, feels like this needs some thought on how hooks intersect with compound actions.

app/components/ChoosePeername.tsx
```diff
@@ -32,13 +36,22 @@ const ChoosePeername: React.FunctionComponent<ChoosePeernameProps> = (props: Cho
...
       setLoading(true)
       resolve()
     })
-      .then(async () => onSave(newPeername))
-      .catch(() => {
-        setLoading(false)
-        setError('some error occured')
+      .then(async () => {
+        setPeername(newPeername)
+          .then((action) => {
```
b5: add a default action type here:

```ts
.then((action = { type: '' }) => {
```

app/components/CommitDetails.tsx
```diff
@@ -1,80 +1,120 @@
...
-  if (meta) return false
-  if (schema) return false
-  return true
-}
-
-export default class CommitDetails extends React.Component<CommitDetailsProps> {
-  state = {
-    selectedCommitPath: ''
-  }
+// const isEmpty = (status: DatasetStatus) => {
```
b5: remove ded code

app/components/Dataset.tsx
```diff
@@ -85,6 +105,25 @@ export default class Dataset extends React.Component<DatasetProps> {
       }
     }
 
+    // make sure that the component we are trying to show actually exists in this version of the dataset
+    // TODO (ramfox): there is a bug here when we try to switch to body, but body hasn't finished fetching yet
```
b5: Let's be sure to file an issue on this

app/components/Metadata.tsx
```diff
@@ -74,6 +76,11 @@ const renderTable = (keys: string[], data: Meta) => {
 
 const Metadata: React.FunctionComponent<MetadataProps> = (props: MetadataProps) => {
   const { meta } = props
+
+  if (!meta) {
```
b5: We should come up with a name for this pattern, and decide if we like it. "no data spinner return" doesn't exactly roll of the tongue, but that's kinda what's going on here. I think an improvement could take one of two approaches:

1. set a rule that makes principle data  _required_ (in this case the `meta` prop). Require that the _parent component must handle loading state_, and can only mount components once data exists. 

2. Functional components must be able to handle not having principle data, and should instead show a "template" of the general component shape, usually they show chunky rectangles instead of text. This is becoming far more popular in modern app development, as users can get a sense of the "shape" of the page before any data has loaded. React Storybook would be _super_ helpful for this one.

This isn't for now, but I think we need to name this pattern and handle it in uniform ways.

app/components/Schema.tsx
```diff
@@ -1,11 +1,16 @@
 import * as React from 'react'
 import { Dataset } from '../models/dataset'
+import SpinnerWithIcon from './chrome/SpinnerWithIcon'
 
 interface SchemaProps {
   schema: Dataset['schema']
 }
 
 const Schema: React.FunctionComponent<SchemaProps> = (props: SchemaProps) => {
+  if (!props.schema) {
```
b5: another example of "no data spinner return"

app/components/chrome/Toast.tsx
```diff
@@ -0,0 +1,30 @@
...
+  text: string
+}
+
+export enum ToastTypes {
+  message = 'message',
+  error = 'error',
+  success = 'success'
+}
+
+const Toast: React.FunctionComponent<ToastProps> = (props: ToastProps) => {
```
b5: 🍻

app/components/modals/CreateDataset.tsx
```diff
@@ -212,13 +215,19 @@ const CreateDataset: React.FunctionComponent<CreateDatasetProps> = ({ onDismisse
...
-    // should fire off action and catch error response
-    // if success, fetchDatatsets
+    error && setError('')
     if (!onSubmit) return
     onSubmit(path, datasetName, bodyFormat)
-      .then(() => onDismissed())
+      .then(() => {
+        fetchMyDatasets()
+        setWorkingDataset('me', datasetName, true)
+          .then(() => {
```
b5: This can be refactored to:

```ts
.then(onDismissed)
```

app/utils/actionType.ts
```diff
@@ -0,0 +1,8 @@
+const getActionType = (type: string): string => {
```
b5: A function named `getActionType` should look like this:
```ts
const getActionType = (action = { type: '' }): string => action.type
```

Rename this to `apiActionType`, and consider accepting the action object instead of a string to make life easier for callers, and consider using the empty string as a default, instead of `'default'`:

```ts
const apiActionType = (action = { type: '' }): string => {
  if (action.type.includes('REQUEST')) return 'request'
  if (action.type.includes('SUCCESS')) return 'success'
  if (action.type.includes('FAILURE')) return 'failure'
  return ''
}

File: app/actions/api.ts [hunk 12]
```diff
@@ -6,6 +6,11 @@ import { Action } from 'redux'
 import { setSelectedListItem } from './selections'
 
 const pageSizeDefault = 15
+const bodyPageSizeDefault = 100
+
+// use NO_ACTION when you need to skip/debounce unneccessary calls to the api
+// it will still register as a success
+export const NO_ACTION: Action = { type: 'NO_ACTION_SUCCESS' }
```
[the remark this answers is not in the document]
ramfox: Thank you. I knew the code smell was high with this one.

app/utils/actionType.ts
ramfox: N2IT

app/components/App.tsx
ramfox: yes! definitely think about it every time I write it...

PR #96, opened 2019-08-07
feat: implement tooltips in dataset component

### LINKED ISSUE #54: [2] feat: Tooltips!
Let's put tool tips everywhere:

* [x] status tab- The status is where you can see the changes that you have made to your dataset since your last commit
* [x] history tab- The history is where you can see all the commits on this dataset
* [x] on meta, body, and schema items
* [x] on link to filesystem button - this will link your dataset to a folder on your computer
* [x] on Show Dataset Files button - shows the path to your dataset folder
* [x] status icons - modified, added, deleted
* [ ] validation status - should give details about # of errors
* [x] current dataset header - add full dataset ref  including peername

chriswhong:
Adds `react-tooltip`.  

ReactTooltip the component is added at the bottom of the render function in `Dataset.tsx`.  Tooltips can quickly be added by adding a `data-tip={'tooltip text'}` attribute to any element.

Closes #54

### DIFF
--- a/app/components/ComponentList.tsx
+++ b/app/components/ComponentList.tsx
@@ -11,23 +11,28 @@ interface FileRowProps {
   status?: string
   selectionType?: ComponentType
   disabled?: boolean
+  tooltip?: string
   onClick?: (type: ComponentType, activeTab: string) => Action
 }
 
 export const FileRow: React.FunctionComponent<FileRowProps> = (props) => {
-  let statusColor
+  let statusColor, statusTooltip
   switch (props.status) {
     case 'modified':
       statusColor = '#cab081'
+      statusTooltip = 'modified'
       break
     case 'add':
       statusColor = '#83d683'
+      statusTooltip = 'added'
       break
     case 'removed':
       statusColor = '#e04f4f'
+      statusTooltip = 'removed'
       break
     default:
       statusColor = 'transparent'
+      statusTooltip = ''
   }
 
   return (
@@ -36,6 +41,7 @@ export const FileRow: React.FunctionComponent<FileRowProps> = (props) => {
         'selected': props.selected,
         'disabled': props.disabled
       })}
+      data-tip={props.tooltip}
       onClick={() => {
         if (props.onClick && props.selectionType && props.name) {
           props.onClick(props.selectionType, props.name)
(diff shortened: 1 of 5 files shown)


app/components/ComponentList.tsx
```diff
@@ -11,23 +11,28 @@ interface FileRowProps {
...
   selectionType?: ComponentType
   disabled?: boolean
+  tooltip?: string
   onClick?: (type: ComponentType, activeTab: string) => Action
 }
 
 export const FileRow: React.FunctionComponent<FileRowProps> = (props) => {
-  let statusColor
+  let statusColor, statusTooltip
   switch (props.status) {
```
b5: Since all of these properties are static, consider using a map constant:

```js
const statusProps = {
  modified: {
    color: '#cab081',
    tooltip: 'modified'
  }
  // ...
}

export const FileRow: React.FunctionComponent<FileRowProps> = (props) => {
   const { statusColor = 'transparent' , statusTooltip = '' } = statusProps[props.status]
  // ...
```

app/components/ComponentList.tsx
```diff
@@ -66,15 +72,18 @@ interface ComponentListProps {
 const components = [
   {
     name: 'meta',
-    displayName: 'Meta'
+    displayName: 'Meta',
+    tooltip: 'View title, description, tags, etc'
```
b5: "View" feels a _little_ like "click for more". maybe just `title, description, tags, etc.`?

app/components/Dataset.tsx
```diff
@@ -62,6 +64,10 @@ export default class Dataset extends React.Component<DatasetProps> {
     setInterval(() => { this.props.fetchWorkingStatus() }, 1000)
   }
 
+  componentDidUpdate () {
+    ReactTooltip.rebuild()
```
b5: would love a comment here explaining why this is necessary. I could see others pulling this pattern into other parts of the app, and knowing why the `rebuild()` call is necessary would help

package.json
```diff
@@ -178,7 +178,8 @@
   },
   "dependencies": {
     "@handsontable/react": "3.0.0",
-    "@types/underscore": "^1.9.2",
+    "@types/react-tooltip": "3.9.3",
+    "@types/underscore": "1.9.2",
```
b5: Also, now that underscore's crept in here as a dep we should know about it and use it. underscore is 🇱🇹 (*LIT* thuania). Off the top of my head, it's debounce func is useful.

app/components/ComponentList.tsx
chriswhong: This is a typescript rabbithole because props.status can be undefined.  Gonna punt on it.

app/components/ComponentList.tsx
chriswhong: To be cleaned up in later iterations when we fine tune all the tooltips

PR #154, opened 2019-09-03
fix(windows): Fix Windows crashes at startup. Improve installer.

dustmop:
Recently added menu item functionality broken Windows at startup. Add some options to the installer so that it behaves more like a normal application installer.


b5:
also, I'm not qualified to say LGTM, FTR.

PR #155, opened 2019-09-03
fix(signin/signup): get session info from response to signin/signup

### LINKED ISSUE #152: Keep profile information from signin and signup responses
If the peername or id is different, we need to clear the current datasets list, clear the selection, and re-fetch the datasets list

ramfox:
Store the session info in Session
If the session is different, reset the selection and pull myDatasets
If we are pulling myDatasets from page 1, added logic that would make sure we are not holding onto the previous list of datasets

closes #152

### DIFF
--- a/app/reducers/index.ts
+++ b/app/reducers/index.ts
@@ -20,20 +20,27 @@ const initialSession: Session = {
 }
 
 const [SESSION_REQ, SESSION_SUCC, SESSION_FAIL] = apiActionTypes('session')
-const [SET_PEERNAME_REQ, SET_PEERNAME_SUCC, SET_PEERNAME_FAIL] = apiActionTypes('set_peername')
+const [SIGNUP_REQ, SIGNUP_SUCC, SIGNUP_FAIL] = apiActionTypes('signup')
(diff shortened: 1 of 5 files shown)


app/reducers/index.ts
```diff
@@ -20,20 +20,27 @@ const initialSession: Session = {
...
 
 const [SESSION_REQ, SESSION_SUCC, SESSION_FAIL] = apiActionTypes('session')
-const [SET_PEERNAME_REQ, SET_PEERNAME_SUCC, SET_PEERNAME_FAIL] = apiActionTypes('set_peername')
+const [SIGNUP_REQ, SIGNUP_SUCC, SIGNUP_FAIL] = apiActionTypes('signup')
+const [SIGNIN_REQ, SIGNIN_SUCC, SIGNIN_FAIL] = apiActionTypes('signin')
 
 const sessionReducer: Reducer = (state = initialSession, action: AnyAction) => { // eslint-disable-line
   switch (action.type) {
-    case SESSION_REQ || SET_PEERNAME_REQ:
+    case SESSION_REQ:
```
b5: we looked at this, and realized it's the only way that works. _boooooooooo javascript_

PR #203, opened 2019-09-09
Various fixes and refactoring

### LINKED ISSUE #52: Commit Components List Should be less wide
Qri_Desktop

The commit components list should be fixed width, and smaller than the current min-width.  We aren't showing any information there other than the component name, so nobody should have a need to make adjust its width

Also remove the subtext since there are no files (make the list items shorter)

### LINKED ISSUE #164: MacOS quiting doesn't work from the dock
right clicking the qri icon in the dock and clicking `quit` does not quit out of qri

You can only quit using the file menu

### LINKED ISSUE #169: Remove tooltip on `Show files` button & fix tooltip on `current dataset`

### LINKED ISSUE #190: UX: New Dataset Modal
- The `Choose` text is too pale
- Help text at the top describing what creating a new dataset will do: "This will create a new dataset from a structured data file. Qri will create a folder for this new dataset."
- Change `Data file` to `Source data file`
- Change `Save Path` to `Dataset folder` ???
- Add `info` icons with tooltip next to each field label

### LINKED ISSUE #201: activeTab state and localstorage mismatch
To reproduce:
- Load a dataset with history, select the history tab
- Switch to a dataset with no history (Desktop will select the status tab, but will not update activeTab in localstorage)
- Refresh
- History tab will be active when user never selected it

### LINKED ISSUE #202: New Dataset Modal allows invalid dataset names
Validate the form input on `CreateDataset`

chriswhong:
- Remove header tooltips, show full `peername/name` in current dataset area (Closes #169)
- Fix unable to quit from MacOS dock (Closes #164)
- Fix activeTab localstorage <-> state mismatch (Closes #201)
- Add dataset name validation in `CreateDataset` modal (Closes #202)
- Remove unnecessary text from commit component list, fix width (Closes #52)

- Adds tooltips, new text, and style tweaks to Create Dataset Modal (Closes #190)
Fullscreen_9_9_19__4_20_PM

### DIFF
--- a/app/components/form/TextInput.tsx
+++ b/app/components/form/TextInput.tsx
@@ -1,7 +1,10 @@
 import * as React from 'react'
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import { faInfoCircle } from '@fortawesome/free-solid-svg-icons'
 
 export interface TextInputProps {
   label?: string
+  labelTooltip?: string
   name: string
   type: string
   value: any
@@ -15,8 +18,22 @@ export interface TextInputProps {
   white?: boolean
 }
 
-const TextInput: React.FunctionComponent<TextInputProps> = ({ label, name, type, value, maxLength, errorText, helpText, showHelpText, onChange, onKeyDown, placeHolder
-}) => {
+const TextInput: React.FunctionComponent<TextInputProps> = (props) => {
+  const {
+    label,
+    labelTooltip,
+    name,
+    type,
+    value,
+    maxLength,
+    errorText,
+    helpText,
+    showHelpText,
+    onChange,
+    onKeyDown,
+    placeHolder
+  } = props
+
   const feedbackColor = errorText ? 'error' : showHelpText && helpText ? 'textMuted' : ''
   const feedback = errorText || (showHelpText &&
     helpText)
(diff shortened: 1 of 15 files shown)


app/components/modals/Modal.tsx
```diff
@@ -238,6 +239,14 @@ const Modal: React.FunctionComponent<ModalProps> = ({ title, dismissable = false
           {children}
         </fieldset>
       </form>
+      <ReactTooltip
```
b5: would be great to get a comment about how we're instantiating a second instance of `ReactTooltip` here because of the interaction between `dialog` and `z-index`

app/components/form/TextInput.tsx
```diff
@@ -15,7 +18,7 @@ export interface TextInputProps {
   white?: boolean
 }
 
-const TextInput: React.FunctionComponent<TextInputProps> = ({ label, name, type, value, maxLength, errorText, helpText, showHelpText, onChange, onKeyDown, placeHolder
+const TextInput: React.FunctionComponent<TextInputProps> = ({ label, labelTooltip, name, type, value, maxLength, errorText, helpText, showHelpText, onChange, onKeyDown, placeHolder
```
b5: Personally I'm a fan of _always_ declaring a `props` argument, moving all destructuring to the first line of the function block. This makes the signature length a little much, especially when blended in with TypeScript declarations.

Maybe we don't "always" do it, but I'd like to see a `props` arg here

app/utils/formValidation.ts
```diff
@@ -27,3 +27,14 @@ export const validatePassword = (password: string) => {
   }
   return null
 }
+
+export const validateDatasetName = (name: string): ValidationError => {
```
b5: This should definitely have a test 😄

chriswhong: 👍

app/components/modals/Modal.tsx
chriswhong: 👍

app/components/form/TextInput.tsx
chriswhong: 👍 agree, this is too much on one line

PR #207, opened 2019-09-09
fix(api, init, add, profile menu): More fixes and refactoring

### LINKED ISSUE #168: bug: User can open their profile app, then click “current dataset”, both menus will be open
need to close profile menu when not focused

### LINKED ISSUE #194: Show meta by default on create/add dataset

### LINKED ISSUE #206: Bug: if dataset returns with error, clear selection
This requires catching the error in `fetchWorkingDataset`

ramfox:
closes #206  
closes #194
closes #168

### DIFF
--- a/app/actions/api.ts
+++ b/app/actions/api.ts
@@ -1,12 +1,13 @@
-import { Action } from 'redux'
+import { Action, AnyAction } from 'redux'
 
-import { CALL_API, ApiAction, ApiActionThunk, chainSuccess } from '../store/api'
+import { CALL_API, ApiAction, ApiActionThunk, chainSuccess, ApiResponseAction } from '../store/api'
 import { DatasetSummary, ComponentStatus, ComponentState, WorkingDataset, ComponentType } from '../models/store'
 import { Dataset, Commit } from '../models/dataset'
 import { openToast } from './ui'
-import { setWorkingDataset, setSelectedListItem } from './selections'
+import { setWorkingDataset, setSelectedListItem, clearSelection, setActiveTab } from './selections'
 
 import { RESET_MY_DATASETS } from '../reducers/myDatasets'
+import getActionType from '../utils/actionType'
 
 const pageSizeDefault = 50
 const bodyPageSizeDefault = 100
@@ -122,6 +123,7 @@ export function fetchMyDatasets (page: number = 1, pageSize: number = pageSizeDe
           state.myDatasets.pageInfo.fetchedAll) {
       return new Promise(resolve => resolve())
     }
+
     const listAction: ApiAction = {
       type: 'list',
       [CALL_API]: {
@@ -169,7 +171,18 @@ export function fetchWorkingDataset (): ApiActionThunk {
       }
     }
 
-    return dispatch(action)
+    const result = (dispatch(action) as unknown) as Promise<AnyAction>
+    return new Promise((resolve, reject) => {
+      result.then(action => {
+        if (getActionType(action) === 'failure') {
+          if (action.payload.err.code === 404 || action.payload.err.code === 500) {
+            dispatch(clearSelection())
+            reject(action)
+          }
+        }
+        resolve(action)
+      })
+    })
   }
 }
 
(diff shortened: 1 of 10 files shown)


app/actions/api.ts
```diff
@@ -169,7 +171,18 @@ export function fetchWorkingDataset (): ApiActionThunk {
       }
     }
 
-    return dispatch(action)
+    const result = (dispatch(action) as unknown) as Promise<AnyAction>
```
b5: Add a comment here explaining what's going on:
```js
// the action being dispatched will be intercepted by api middleware and return a promise
// typescript requires we first cast to unknown to drop the stated return type of the dispatch function
const result = (dispatch(action) as unknown) as Promise<AnyAction>
```

app/actions/api.ts
```diff
@@ -169,7 +171,18 @@ export function fetchWorkingDataset (): ApiActionThunk {
...
+    return new Promise((resolve, reject) => {
+      result.then(action => {
+        if (getActionType(action) === 'failure') {
+          if (action.payload.err.code === 404 || action.payload.err.code === 500) {
+            dispatch(clearSelection())
```
b5: add a comment here explaining what's happening:
```js
// if the working dataset isn't found, we have dataset selection that no longer exists. clear the selection.
dispatch(clearSelection())
```

app/actions/api.ts
```diff
@@ -441,6 +454,12 @@ export function addDatasetAndFetch (peername: string, name: string): ApiActionTh
     try {
       response = await addDataset(peername, name)(dispatch, getState)
       response = await whenOk(fetchMyDatasets())(response)
+      const action = response as ApiResponseAction
+      const { data } = action.payload
```
b5: If we're not using a default value for data (because we _know_ data will be present), there's no need for this destructuring.

Refactor to just:
```js
const { isLinked, published } = action.payload.data.find((dataset: DatasetSummary) => dataset.name === name && dataset.peername === peername)
```

app/actions/api.ts
```diff
@@ -475,6 +494,12 @@ export function initDatasetAndFetch (sourcebodypath: string, name: string, dir:
     try {
       response = await initDataset(sourcebodypath, name, dir, mkdir)(dispatch, getState)
       response = await whenOk(fetchMyDatasets())(response)
+      const action = response as ApiResponseAction
+      const { data } = action.payload
```
b5: Same as above, no need to destructure `data`

app/components/chrome/HeaderColumnButtonDropdown.tsx
```diff
@@ -12,10 +12,30 @@ const HeaderColumnButtonDropdown: React.FunctionComponent<HeaderColumnButtonDrop
...
+    if (dropdown.current !== null && dropdown.current.contains(e.target)) {
+      // inside click
+      return
+    }
+    // outside click
+    setShowMenu(false)
+  }
+
+  React.useEffect(() => {
+    document.addEventListener('mousedown', (e) => handleClick(e))
```
b5: any reason not to use the [`click` event](https://caniuse.com/#search=click%20event) instead of `mousedown`? mousedown has funny implications: if you click and hold, it'll fire the action right away.

app/store/api.ts
```diff
@@ -56,6 +56,18 @@ interface ApiQuery {
...
 }
 
+interface ApiError {
+  code: number
+  message?: string
+}
+
+export interface ApiResponseAction {
+  type: string
+  payload: {
```
b5: consider refactoring to:
```ts
payload: Record<string,any>
```

This will require `payload` be an object with no functions.

app/store/api.ts
```diff
@@ -56,6 +56,18 @@ interface ApiQuery {
   [key: string]: string
 }
 
+interface ApiError {
```
b5: pretend this is go. Let's get a comment defining the struct, a comment on `code` clarifying that it'll be an HTTP response code, and a comment on message "error message returned from qri"

app/store/api.ts
```diff
@@ -56,6 +56,18 @@ interface ApiQuery {
...
+  code: number
+  message?: string
+}
+
+export interface ApiResponseAction {
```
b5: ditto to linting requirements:
```ts
// ApiResponseActions are emitted as side effects by api middleware. They come in "success" and "error" variations
export ApiResponseAction
```

app/components/chrome/HeaderColumnButtonDropdown.tsx
ramfox: no reason, adjusting now

ramfox:
Your review made me realize I haven't been doing my duty in commenting code! Resolving to have better commenting habits.

PR #210, opened 2019-09-10
feat: validate commit form

### LINKED ISSUE #193: Disable commit button unless something is modified AND there is a commit message
Also check that there are no errors in status

### LINKED ISSUE #195: Nudge new users to commit
When user is looking at a new dataset with no commits, add a pane to nudge them towards a commit (You’re ready to create your first commit of this dataset.  Review meta, verify schema, check out the body, etc)

chriswhong:
Adds better validation to the commit form (Closes #193)
Adds a permanently hideable call to action above the commit form on datasets with no history (Closes #195 )

- Uses `useState` to manage the user's input locally, removes an action and reducer for storing the commit message in the state tree
- Validates and disables submit button on changes to `title`, `message`, or working dataset `status`
- Adds validation rules: 
  - commit tile must be > 3 characters
  - component statuses must include at least one 'modified'
  - component statuses may not include anything matching `/error/g`

- The commit button is now blue, matching the commit button on our modal forms.

- Refactors `Dataset` & `DatasetContainer`, moving a lot of props and actions down to the new `DatasetSidebarContainer`.


app/containers/DatasetSidebarContainer.tsx
```diff
@@ -0,0 +1,57 @@
+import { connect } from 'react-redux'
```
b5: pumped to see this turned into a container!

app/components/SaveForm.tsx
```diff
@@ -1,66 +1,90 @@
...
+
+  const [isValid, setIsValid] = React.useState(false)
+  const [title, setTitle] = React.useState('')
+  const [message, setMessage] = React.useState('')
+
+  React.useEffect(() => {
+    // validate form -AND- make sure dataset status is in a commitable state
+    let valid = true
+    // commit message (title) must be more than 3 characters
+    if (title.length < 4) valid = false
```
b5: Despite being trivial, this is business logic. To keep this component stupid, we should move it to a function in `app/utils/formValidation.js` and add a test. Over time we may get more sophisticated with title & message validation rules.

chriswhong: 👍

PR #213, opened 2019-09-10
refactor(selections): all actions that should effect selection are caught in the reducer

### LINKED ISSUE #197: After creating a dataset, `unlinked dataset` page shown erroneously
Due to a bug surrounding peername

### LINKED ISSUE #212: Catch actions in the `selection` reducer whose responses should effect `selection`
- `DATASET_FAILURE`: should clear peername, name, islinked, and published, activetab, component, commit, commitComponent (everything)
- `DATASET_SUCCESS`: should set selected component, if current selected component is empty or doesn't exist on this dataset
- `COMMIT_SUCCESS`: should set selected commit component if empty or current selected commit component doesn't exist on this version
- `ADD_SUCCESS`: should change peername, name, lisLinked, published, activeTab: history
- `INIT_SUCCESS` should change peername, name, isLinked, published, activeTab: status
- `PUBLISH_SUCCESS`: should change published to true
- `UNPUBLISH_SUCCESS`: should change published to false
- `SIGNIN_SUCCESS`/`SIGNUP_SUCCESS`: peername, name, islinked, and published, activetab, component, commit, commitComponent (everything)

- refactor `routes.tsx` to remove references to `clear selection` and `setWorkingDataset`
- refactor `CommitDetails` to remove effect that `setSelectedListItem` for which commit component to show
- refactor `Dataset` to remove effect that `seSelectedListItem` from which component to show

ramfox:
closes #212 
closes #197 

Right now the reducer changes are in place, but I still need to remove the now-erroneous calls to the selection actions

Added `chooseDefaultComponent` and accompanying tests to keep logic for choosing a default component in one location


app/components/Dataset.tsx
```diff
@@ -161,27 +165,6 @@ export default class Dataset extends React.Component<DatasetProps> {
       }
     }
 
-    if (!_.isEmpty(nextProps.workingDataset.status)) {
-      // make sure that the component we are trying to show actually exists in this version of the dataset
-      // TODO (ramfox): there is a bug here when we try to switch to body, but body hasn't finished fetching yet
```
b5: damn. @ramfox you're confident this switch addresses this comment? If so, we're definitely on the right track

app/components/Dataset.tsx
```diff
@@ -301,7 +284,7 @@ export default class Dataset extends React.Component<DatasetProps> {
               </div>
               <div className='header-column-text'>
                 <div className="label">{name ? 'Current Dataset' : 'Choose a Dataset'}</div>
-                <div className="name">{peername}/{name}</div>
+                <div className="name">{peername !== '' && name !== '' ? `${peername}/${name}` : ''}</div>
```
b5: not a massive fan of doing logic work within templates. can we get a computed `const` before we hit the `JSX` section of render?
```tsx
const alias = (peername != '' && name != '') ? `${peername}/${name}` : ''
// ...
<div className="name">{alias}</div>
```

app/reducers/selections.ts
```diff
@@ -19,11 +20,20 @@ const initialState: Selections = {
...
 const [, ADD_SUCC] = apiActionTypes('add')
+const [, INIT_SUCC] = apiActionTypes('init')
+const [, DATASET_SUCC, DATASET_FAIL] = apiActionTypes('dataset')
+const [, COMMIT_SUCC] = apiActionTypes('commit')
+const [, PUBLISH_SUCC] = apiActionTypes('publish')
+const [, UNPUBLISH_SUCC] = apiActionTypes('unpublish')
+const [, SIGNIN_SUCC] = apiActionTypes('signin')
+const [, SIGNUP_SUCC] = apiActionTypes('signup')
 
 export default (state = initialState, action: AnyAction) => {
```
b5: any chance we can get a skeletal reducer test for this? Feels like it'd be great even to just test the initial state and add a bunch of TODOs for missing tests, but also to carry our patterns for reducer tests over from the frontend repo

app/components/Dataset.tsx
ramfox: yep! part of the issue is that `body` doesn't exist on dataset until it's triggered by actually fetching the body.

The solution is to also check for `bodyPath` (which lets us know that body exists on this version) and be okay with the loading screen while it fetches the body.

app/components/Dataset.tsx
ramfox: can do :)

ramfox:
Wrote tests that pass locally!
we don't have ci set up yet for this repo, got the verbal okay from @b5 to merge

PR #225, opened 2019-09-12
Refactor(*): stabilizing app

### LINKED ISSUE #191: Wrap myDatasets and body requests in load functions that check if we have that page before making the request
Cuts down on duplicate requests.

Check every place we are fetching data, especially paginated requests.

### LINKED ISSUE #196: When scrolling / resizing body the Qri icon shows up while loading
Investigate whether the behavior is different when scrolling the fsi linked filesystem body vs the body in the commit.

### LINKED ISSUE #223: refactor(map functions): create file that contains all map functions used in api
When we get a response, we have a chance to map the data we received to a structure that the reducer is expected.

Often, these map functions are being repeated in may api calls. When something changes, or a bug is found in one function, it is hard to be sure that we are changing them everywhere they need to be changed.

Let's create a separate function for each kind of mapping, put them all in one file, and use the same function each time we need to map something the same way:

1) mapRecord - casts the incoming data as Record<string, string>
2) mapDataset - casts the incoming data as a type Dataset. Adds `isLinked` param derived from fsiPath. This will make it easier if we want to deprecate isLinked in favor of the truthy fsiPath string
3) mapDatasetSummary - pulls apart each element in the `myDatasets` response, marshals them each into a `DatasetSummary`, returning a `DatasetSummary[]`
4) mapStatus - pulls apart each element in the response to marshal into a `ComponentStatus`, returns a `ComponentStatus[]`
5) mapHistory - pulls apart each element in the history response and marshals them into a `Commit`, returns a `Commit[]`

Will also make it easier to test in the future

ramfox:
closes #223 
closes #191 
closes #196 

This fixes our loading bugs, however, the app doesn't feel as snappy. Trade off for knowing exactly what is happening when and why and at what time. We can build on this!

### DIFF
--- a/app/actions/api.ts
+++ b/app/actions/api.ts
@@ -1,10 +1,16 @@
 import { Action, AnyAction } from 'redux'
 
 import { CALL_API, ApiAction, ApiActionThunk, chainSuccess, ApiResponseAction } from '../store/api'
-import { DatasetSummary, ComponentStatus, ComponentState, WorkingDataset, ComponentType } from '../models/store'
-import { Dataset, Commit } from '../models/dataset'
+import { DatasetSummary, WorkingDataset, ComponentType } from '../models/store'
 import { openToast } from './ui'
 import { setWorkingDataset, setSelectedListItem, clearSelection, setActiveTab } from './selections'
+import {
+  mapDataset,
+  mapRecord,
+  mapDatasetSummary,
+  mapStatus,
+  mapHistory
+} from './mappingFuncs'
 
 import { RESET_MY_DATASETS } from '../reducers/myDatasets'
 import getActionType from '../utils/actionType'
@@ -19,9 +25,7 @@ export function pingApi (): ApiActionThunk {
       [CALL_API]: {
         endpoint: 'health',
         method: 'GET',
-        map: (data: Record<string, string>): any => { //eslint-disable-line
-          return data
-        }
+        map: mapRecord
       }
     }
     return dispatch(pingAction)
@@ -70,9 +74,7 @@ export function fetchModifiedComponents (): ApiActionThunk {
           peername,
           name
         },
-        map: (data: Record<string, string>): Dataset => {
-          return data as Dataset
-        }
+        map: mapDataset
       }
     }
     response = await dispatch(resetComponents)
@@ -92,9 +94,7 @@ export function fetchModifiedComponents (): ApiActionThunk {
           name,
           path
         },
-        map: (data: Record<string, string>): Dataset => {
-          return data as Dataset
-        }
+        map: mapDataset
       }
     }
     response = await dispatch(resetBody)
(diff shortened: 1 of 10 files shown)


app/actions/api.ts
```diff
@@ -103,46 +102,40 @@ export function fetchModifiedComponents (): ApiActionThunk {
   }
 }
 
-// clears the dataset list
-function resetMyDatasets (): Action {
-  return {
-    type: RESET_MY_DATASETS
+function actionWithPagination (invalidatePagination: boolean, page: number, pageInfo: PageInfo): ActionWithPaginationRes {
```
chriswhong: 🔥

app/components/Body.tsx
```diff
@@ -40,6 +40,9 @@ const Body: React.FunctionComponent<BodyProps> = ({ value, pageInfo, headers, st
     if (pageInfo.page === 1 && pageInfo.isFetching) {
       isLoadingFirstPageRef.current = true
     }
+    if (isLoadingFirstPage && isLoadingFirstPageRef.current === false) {
```
chriswhong: I'm not following on why we need a `ref` here instead of just another boolean in state.

I think some comments on the flow that's happening here would be very helpful.  Why do we need a ref?  Why do we need a 150ms timeout? 
```
  const [isLoadingFirstPage, setIsLoadingFirstPage] = React.useState(false)
  const isLoadingFirstPageRef = React.useRef(pageInfo.page === 1 && pageInfo.isFetching)

  const loadingTimeout = setTimeout(() => {
    if (isLoadingFirstPageRef.current) {
      setIsLoadingFirstPage(true)
    }
    clearTimeout(loadingTimeout)
  }, 250)

  React.useEffect(() => {
    if (pageInfo.page === 1 && pageInfo.isFetching) {
      isLoadingFirstPageRef.current = true
    }
    if (isLoadingFirstPage && isLoadingFirstPageRef.current === false) {
      setIsLoadingFirstPage(false)
    }
  }, [pageInfo.page, pageInfo.isFetching])
```

b5:
@chriswhong I took a look at your review, and pushed a commit to this branch that I think addresses your concerns.

In essence, @ramfox's code was trying to maintain sync between the local `useState` and other changes this PR makes that were delivering the same details as passed in props. @ramfox needed `useRef`'s imperative properties to make this happen outside of render cycles. My guess is she was trying to avoid breaking code she didn't write (the principle `useState`)

I, on the other hand, am a jerk who doesn't care about breaking other people's code, so I ripped the useState out entirely in favor of just relying on passed-in props. Less code, no timeout, works great.

ramfox:
Hey, timeout was to prevent loading flashing! Basically a stand-in for suspense until suspense does what it says it will. Basically: prevent the loading screen for 150ms until we are sure we are loading long enough to merit showing the user.

It was confusing and should have been documented!

PR #229, opened 2019-09-17
refactor: standardize fsiPath, isLinked, linkpath, simplify selections

### LINKED ISSUE #228: New Dataset flow lands user in awkward state
Observed during today's Qri dev call, when a user finishes the new/create dataset workflow, they land on a state where:
- status tab is selected, but all components are grayed out (body and meta should be visible after finishing the new dataset workflow)
- history tab is clickable, but shows the qri logo and a loading spinner (tab should be disabled if there is no history)

chriswhong:
Closes #228, a bug where we observed the create dataset flow resulting in an awkward state requiring a reload.

On investigation, the the error was due to API calls with `?fsi=undefined`.

This PR refactors how we store state for a dataset's linked status, as follows:
- We had 3 different names for linked status: `isLinked` (boolean), `linkpath` (string), and `fsiPath`(string).  These are now fully standardized as `fsiPath`, which is what the API returns (this simplifies mapping functions).  `isLinked` is used only in the `Dataset` component as a local convenience variable.
- `isLinked` and `published` are now removed from `selections`.  They should never have been there as they aren't user selections that need to be persisted, they are simply the state of the `workingDataset` (this also means they are not persisted in localstorage)

`fsi` query params are now being applied correctly, and both unlinked and linked datasets are rendering as expected.

### DIFF
--- a/app/actions/api.ts
+++ b/app/actions/api.ts
@@ -1,7 +1,7 @@
 import { Action, AnyAction } from 'redux'
 
 import { CALL_API, ApiAction, ApiActionThunk, chainSuccess, ApiResponseAction } from '../store/api'
-import { DatasetSummary, ComponentType, PageInfo } from '../models/store'
+import { DatasetSummary, ComponentType, PageInfo, MyDatasets } from '../models/store'
 import { openToast } from './ui'
 import { setWorkingDataset, setSelectedListItem, clearSelection, setActiveTab } from './selections'
 import {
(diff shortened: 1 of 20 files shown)


ramfox:
repeating my commit message here to outline why we are making the `fetchWorkingDataset` changes:

When we are trying to fetchWorkingDataset, we are gathering information for the dataset status. If the dataset is NOT fsi linked, we don't need any other information besides the peername and dataset name to get all the information we need to populate the history page.

So, let's always assume that when we are trying to `fetchWorkingDataset` we believe that `fsi=true`. We've added the 422 error code on the backend to let us know that the dataset is not linked. (which is different from 404 does not exist, since the dataset does exist in qri, it just isn't linked to the file system)

This commit makes all calls using the `fetchWorkingDataset` action to assume `fsi=true`

When we `DATASET_REQUEST`, we take the peername and name from the action and add it to the state tree

If the `DATASET_FAIL` action emits from this dispatch, but the action errored with a 422 status code, do no clear the full working dataset, allow the peername and name to remain, and ensure that we are marking `fsiPath=''`

In the `fetchWorkingDatasetDetails`, we need to catch the emitted error from `fetchMyDataset`, if it's the 422 error we still need to `fetchWorkingHistory`

chriswhong:
@ramfox Looks great! I smoke tested locally and noticed we lost the line that sets `mtime` in the status reducer (I think when we moved stuff to `mappingfuncs`), and the status polling interval was too long.  I amended both of those changes and I think we are in business.

PR #233, opened 2019-09-18
test(circleci): add circleci config

ramfox:


.circleci/config.yml
```diff
@@ -0,0 +1,36 @@
...
+      - run: 
+          name: yarn install
+          command: yarn install
+      - save_cache:
+          paths:
+            - node_modules
+          key: dependency-cache-{{ checksum "yarn.lock" }}
+      - run:
+          name: lint with eslint
+          command: yarn lint
```
chriswhong: I think we should avoid automatically running a linter with the `--fix` option, which could potentially modify the code _after_ it lands on the testing environment.  I propose that `yarn lint` with the fix flag should only ever be run manually in development, to help clean things up before committing code, and we should add a proper test to `yarn test` that runs eslint without the --fix flag and will have a non-zero exit code on linting errors.  [Here's an example of running eslint as its own mocha test](https://github.com/NYCPlanning/labs-zap-api/blob/develop/test/eslint/eslint-test.js) and a quick search came up with many options to do the same in jest.

ramfox: Oh yes good catch! Def should not be running this with `--fix`

ramfox: Okay, so removed `--fix` and added linting errors on purpose. Then ran ci locally, and it exited with an error!

PR #267, opened 2019-09-24
fix(bootstrap, backend version, defaultPollInterval, pageInfo): lots of fixes

### LINKED ISSUE #189: “About Qri Desktop” in app menu should show backend version #

### LINKED ISSUE #250: refactor: bootstrapping the app
When we open or refresh the app:

create a compound api action that 
- asks for the session
- if successful, calls `fetchMyDatasets`
- if successful, calls `fetchWorkingDatasetDetails`

`App` component should call this action on `ComponentDidMount`

- catch any `unauthorized` errors in `apiMiddleware` (same place we catch `FAILED_TO_FETCH` errors), emit error, catch error in UI reducer to set `ui.qriCloudAuthenticated` false

### LINKED ISSUE #260: bug: cannot read pageInfo of undefined
pretty sure this happens for history lists but will double check

effect is that history list and contents always in `isFetching: true` state. Makes me think either the `HISTORY_SUCC` or `HISTORY_FAIL` is pulling pageInfo from the wrong place.

### LINKED ISSUE #261: refactor: add `defaultPollingMs` constant
use this constant to set the health check interval in  `App` and status check interval in `Dataset`

ramfox:
closes #261 
closes #260 
closes #189 
closes #250

### DIFF
--- a/app/main.development.js
+++ b/app/main.development.js
@@ -2,6 +2,11 @@ const { app, BrowserWindow, Menu, shell, ipcMain } = require('electron')
 const { autoUpdater } = require('electron-updater')
 const { BackendProcess } = require('./backend')
 
+// versions
+// must be manually updated for now
+const desktopVersion = '0.1.0-rc1'
+const backendVersion = '0.9.0-alpha'
+
 let menu
 let template
 let mainWindow = null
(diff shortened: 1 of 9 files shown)


app/main.development.js
```diff
@@ -2,6 +2,11 @@ const { app, BrowserWindow, Menu, shell, ipcMain } = require('electron')
 const { autoUpdater } = require('electron-updater')
 const { BackendProcess } = require('./backend')
 
+// versions
+// must be manually updated for now
+const desktopVersion = '0.1.0-rc1'
+const backendVersion = '0.9.0-alpha'
```
b5: eventually this should be populated by a webpack build constant

PR #280, opened 2019-09-25
fix(*): Lots o fixes

### LINKED ISSUE #272: bug: when clicking '+ new dataset' if no path specified already, crash
looks like we need a default path if we haven't set one already

### LINKED ISSUE #274: bug: after commit, recent save did not show up in dataset history
after commit, we need to re-fetch history and reset pagination: `fetchHistory(-1)`

### LINKED ISSUE #275: bug: when switching to history tab, there is a flash before the history list is in place

### LINKED ISSUE #276: refactor: "New Dataset" => "Create New", "Add Dataset" => "Add from Network"

### LINKED ISSUE #277: bug: if you remove meta.json, you can't get to the meta editor
We should always allow `meta` to be "clickable"

### LINKED ISSUE #279: bug: if a component file does not exist, then does, it doesn't trigger re-fetch of dataset

### LINKED ISSUE #281: refactor(body): if csv or excel, always render as a table
if it's json & depth === 2, show json viewer

ramfox:
closes #274 
closes #279 
closes #272 
closes #277 
closes #281
closes #275 
closes #276 

also:
- final fix on fetching session, keep app in loading state until session returns
- simplify process on reseting components, fixes bugs with saving after meta editing
- we keep going back and forth on this, but I now don't think that dataset_success should write over the body
- refactor Dataset to remove `getDerivedPropsFromState` to `ComponentDidUpdate`
- refactored how we set commits in selection after we fetchHistory for the first time

### DIFF
--- a/app/actions/api.ts
+++ b/app/actions/api.ts
@@ -46,7 +46,6 @@ export function fetchWorkingDatasetDetails (): ApiActionThunk {
     const whenOk = chainSuccess(dispatch, getState)
     let response: AnyAction
     response = await fetchWorkingDataset()(dispatch, getState)
-
     // if the response returned in error, check the error
     // if it's 422, it means the dataset exists, it's just not linked to the filesystem
     if (getActionType(response) === 'failure') {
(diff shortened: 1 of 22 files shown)


app/actions/api.ts
```diff
@@ -124,7 +115,8 @@ export function fetchMyDatasets (page: number = 1, pageSize: number = pageSizeDe
     const state = getState()
     const { page: confirmedPage, doNotFetch } = actionWithPagination(page, state.myDatasets.pageInfo)
 
-    if (doNotFetch) return new Promise(resolve => resolve())
+    // we need to emit a 'success' type, or our chainSuccess functions will fail
+    if (doNotFetch) return new Promise(resolve => resolve({ type: 'SUCCESS' }))
```
b5: This is fine for now. Later we should expand the definition of what `chainSuccess` to allow empty resolves and do the right thing when resolved with no arguments

PR #293, opened 2019-10-02
feat(export): export a specific version of a dataset

ramfox:
uses `electron-dl` package to manage downloads

* [x] create modal
* [x] save folder input (default to platform specific downloads folder)
* [x] `export` button that emits the `ipcRender.Send('export'...)` func

### DIFF
--- a/app/components/modals/ExportVersion.tsx
+++ b/app/components/modals/ExportVersion.tsx
@@ -1,0 +1,134 @@
+import * as React from 'react'
+import { Action } from 'redux'
+import { remote, ipcRenderer } from 'electron'
+
+import Modal from './Modal'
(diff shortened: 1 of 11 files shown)


app/components/modals/ExportVersion.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+    if (!directory) {
+      return
+    }
+
+    const selectedPath = directory[0]
+
+    setExportPath(selectedPath)
+  }
+  console.log(timestamp)
```
chriswhong: console!

ramfox: good looks

Issue #308, opened 2019-10-25
Body view refactor


chriswhong:
Three types of bodies that have different rendering concerns:

1. Tabular (csv, json array of objects with depth of 2, xlsx)
2. JSON Array of Objects with depth greater than 2 (related to #302)
3. JSON Object (A geojson FeatureCollection is an example of this)

Preliminary work has been done on the branch `json-performance` to use `react-window` for higher performance scrolling of item 2 above. 

HandsonTable is working, but feels slow/unresponsive at times.  Adding a page of data makes the app crawl, presumable due to the calculations necessary to determine column width, etc

An alternative approach may be to also use `react-window` but use html tables to offload the layout calculations.

#1 and #2 should be designed and optimized to provide a smooth UX for pagination (#1 does this now with a toast, #2 does not.  We should avoid the toast and go with some other loading state indicator when pagination is firing)

We can de-prioritize # 3, and use codemirror or just `JSON.stringify()` to show everything (this assumes no pagination)

b5:
@ramfox do you think #313 made enough progress to call this closed?

ramfox:
yes! closed by #313

PR #313, opened 2019-10-31
feat: two-way body pagination, json display

chriswhong:
- To improve performance, replaces handsontable with a custom html table layout for csv dataset body views
- Adds two-way pagination to table bodies, fetching 50 rows at a time and maintaining scroll position
- Adds a `BodyJson` component for rendering JSON bodies which does not paginate
  - If top level json element has more than 50 items, displays a warning that this is a preview only
  - `react-json-view` is configured with `collapsed=2` so deeply nested data are not rendered (but can be expanded manually), improving performance.

Qri_Desktop_and_desktop_—_Electron_Helper_◂_node____nvm_versions_node_v12_1_0_bin_yarn_dev_—_113×41_and_selections_test_ts_—___qri_desktop


chriswhong:
Closes #308

ramfox:
This is exciting!

I'm actively looking for bugs and so far have found only two situations that mess up our new table, both happening when there is no structure or there are structure changes
- If you have no structure, and then make a change to the structure page, we should re-fetch the body
- If you have no structure/schema, the floating header is blank. So instead, we should remove the header entirely.

after that we should be good to merge!

PR #325, opened 2019-11-14
feat(storybook,Stat): initial string Stat component with story

b5:
Given that @ramfox is working on the sliding-panel stuff, and I want to work on starts charts, this is a good chance to bring storybook into the mix.

I've also started building out a basic Stat component that displays frequencies if they exist. There's a *lot* of variability to stats frequency counts, so it's worth trying out different stats output for visual edge cases.

I ended up roughing-in https://nivo.rocks as a charting library. It's configuration-as-docs is an incredible concept.

start the storybook with `yarn storybook`.


ramfox:
LFTM

PR #330, opened 2019-11-15
feat(transform): initial support for transform display

b5:
this adds _very_ basic support for displaying transforms within desktop. Meant mainly as a thing we can iterate on. I've also added a story around the `Code` component, hoping we can make that look not terrible.


app/components/ComponentList.tsx
```diff
@@ -145,7 +162,7 @@ const ComponentList: React.FunctionComponent<ComponentListProps> = (props: Compo
           Dataset Components
       </div>
       {
-        components.map(({ name, displayName, tooltip, icon }) => {
+        components.filter(removeHiddenComponents(status)).map(({ name, displayName, tooltip, icon }) => {
```
b5: one thing to note, I've added a filter here to _not_ display the transform component if it's missing.

Transforms differ from other components: they are _optional_, and users should be able to use Qri without ever interacting with the (potentially scary) transform component if they don't want to.

PR #357, opened 2019-12-04
feat: add dataset rename

chriswhong:
Adds "edit-in-place" UI for renaming datasets when in the dataset view.
Provides visual feedback when the proposed name is invalid, provides success or error toasts after backend response.
Editing happens in a new component `<DatasetReference>`, which could theoretically be re-used to rename datasets in the Collection view.

Screen Shot 2019-12-04 at 1 45 56 PM

### DIFF
--- a/app/components/DatasetReference.tsx
+++ b/app/components/DatasetReference.tsx
@@ -1,0 +1,112 @@
+// a component that displays the dataset reference including edit-in-place UI
+// for dataset rename
+import * as React from 'react'
+import classNames from 'classnames'
+
+import { ApiActionThunk } from '../store/api'
+import { validateDatasetName } from '../utils/formValidation'
+
+interface DatasetReferenceProps {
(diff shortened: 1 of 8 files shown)


app/components/DatasetReference.tsx
```diff
@@ -0,0 +1,112 @@
...
+    }
+  }, [ name, newName ])
+
+  const handleInputChange = (e: any) => {
+    let { value } = e.target
+    setInvalid(validateDatasetName(value))
+    setNewName(value)
+  }
+
+  // when the input is focused, set the cursor to the left, and scroll all the way to the left
```
dustmop: Is this still happening? Or is the comment out of date?

chriswhong: Good catch, it's only scrolling left, not setting the cursor position anymore.  I updated the comment to reflect the code.

PR #364, opened 2019-12-10
refactor: clean up state loading flow

### LINKED ISSUE #363: Audit api calls and rendering when toggling between dataset and collection
- There is double calling happening when viewing commits (I think only on first load, need to investigate)
- There are inefficient requests happening when toggling between dataset and collection view.  These should be refactored to persist between the route changes, and not to unmount and mount when toggling.

This issue consists of auditing the flow of api calls and state updates and refactoring to an optimized state.

chriswhong:
Closes #363 

Refactors flow of data loading API calls to prevent unnecessary calls when navigating around the app.  
- Listener for workingDataset fetch is moved up to `App` level
- Ensured that renaming a dataset would not cause a fetch (it used to because the name and peername had changed)
- Switching between status and history tabs no longer causes any fetching
- Switching between Collection and Dataset view no longer causes any fetching

### DIFF
--- a/app/actions/api.ts
+++ b/app/actions/api.ts
@@ -51,6 +51,7 @@ export function fetchWorkingDatasetDetails (): ApiActionThunk {
     response = await fetchWorkingDataset()(dispatch, getState)
     // if the response returned in error, check the error
     // if it's 422, it means the dataset exists, it's just not linked to the filesystem
+    // we still need the working dataset because it contains the history
     if (getActionType(response) === 'failure') {
       if (response.payload.err.code !== 422) {
         return response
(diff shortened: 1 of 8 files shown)


app/components/App.tsx
```diff
@@ -133,6 +134,15 @@ class App extends React.Component<AppProps, AppState> {
       this.props.bootstrap()
     }
 
+    // listen for changes to current dataset, fetch data on change
+    if ((this.props.selections.peername !== prevProps.selections.peername) || (this.props.selections.name !== prevProps.selections.name)) {
+      // if the paths are the same, don't get data (dataset was renamed)
+      if (this.props.workingDataset.path === prevProps.workingDataset.path) return
+      console.log('fetching again')
```
ramfox: remove console.log

app/actions/api.ts
```diff
@@ -139,21 +140,28 @@ export function fetchMyDatasets (page: number = 1, pageSize: number = pageSizeDe
...
-    const { selections } = getState()
+    const { selections, myDatasets } = getState()
     const { peername, name } = selections
 
     if (peername === '' || name === '') {
       return Promise.reject(new Error('no peername or name selected'))
     }
 
+    // look up the peername + name in myDatasets to determine whether it is FSI linked
+    const dataset = myDatasets.value.find((d) => (d.peername === peername) && (d.name === name))
```
ramfox: this is going to break when the dataset list is very long and paginated :(

This is why we had to guess the first time we fetch.

app/components/App.tsx
chriswhong: removed

app/actions/api.ts
chriswhong: I think we have been operating on the assumption of not doing pagination for the dataset list.  I'd love to talk this through, but I don't think it's unreasonable to always have the full dataset list in state, and keep it in sync.

ramfox: Cool, let's definitely talk. Basically the only reason it would be bad is if the responses from the backend stop being snappy, right now we reset the dataset list alot and if those calls and responses end up being slow it could make the app less fun. We might need to do some tests to see how large the response can be without being a burden.

PR #375, opened 2019-12-13
fix(App, Dataset, Remove): fix eventListener leak and clearing selection after a remove

### LINKED ISSUE #368: refactor: remove checkout of files when adding dataset from cloud
Like the streamlined 'create a new dataset' flow, let's remove the 'checkout' process when you add a dataset from cloud. 

For now, we are going to disable the ability to 'checkout' a dataset that is not in their namespace. Until we get branching correct, we aren't going to give users the option to foot-gun themselves

Edit: After a talk in office, we are going to also remove the folder picker from the "Create a new" dataset modal as well.

* [x] remove folder picker from create dataset modal
* [x] remove folder picker from add dataset modal
* [x] remove ability to 'Checkout' a dataset that isn't in your namespace (remove checkout button, remove any text with a link to checkout, disable checkout if there is a shortcut)
* [x] change checkout page to describe that this dataset isn't yours and if you want to play around with it do these steps (Go to the version you want to work on and right click, click export, unzip the file, drag the body file over to the app to create a new dataset, checkout that dataset, if you want to keep the meta, readme, etc, add those files from the unzipped folder to the new dataset folder. Qri will automatically detect it.)

### LINKED ISSUE #369: refactor: drop user into history after a commit
When you save a commit, we should drop the user into the history tab at the new commit.

### LINKED ISSUE #371: Peername and name not populating in checkout modal
Check out modal title just says "Link /", peername and name are missing.
Also, the text should be reworded to say "Checkout" instead of "Link"
Screen Shot 2019-12-13 at 10 39 27 AM

### LINKED ISSUE #372: bug(rename): enforce the same naming rules as backend
- enforce the same naming rules as the backend (specifically not starting the dataset with a number)
- auto-lower case when a user types the dataset name

### LINKED ISSUE #373: MaxListenersExceededWarning in console (IPC listeners)
To reproduce:  Toggle between Dataset and Collection view 11 times.  This is happening because ipc event listeners are not being turned off when components unmount.  

`ipcRenderer.on('publish-unpublish-dataset', this.publishUnpublishDataset)`

There is a `removeListener()` method and a `removeAllListeners()` method on `ipcRenderer` https://electronjs.org/docs/api/ipc-renderer#ipcrendererremovelistenerchannel-listener

Qri_Desktop

### LINKED ISSUE #374: bug(dataset): When you remove a dataset, clear the selection
Right now, when you remove a dataset, and then switch back to the dataset screen, that dataset is still selected, and we fire off a fetch for that dataset.

ramfox:
The `ipcRenderer.on()` functions that we use to hook into the electron menu are event listeners. We haven't been handling the removal of these listeners in the `App` and `Dataset` components. For each listener we add in `componentDidMount` we need to remove it in `componentDidUnmount`.

This commit also fixes some styling on the Dataset component when a dataset is not `checkedout`, it clears the selection once you've successfully removed a dataset, and also adjusts the no dataset selected state to include navigation that will send you back to the collections page.

closes #368 - remove auto checkout from Create and Add modals/flow, also remove ability to checkout a dataset that is not in your namespace, until we come up with an explicit plan (remove ability for Desktop users to foot gun themselves)
closes #369 - after a commit, drop user into history to view the commit
closes #371 - adjust checkout modal _note: could not replicate the bug in the issue_
closes #372 - bring desktop naming rules inline with backend
closes #373 - fix event listeners leak
closes #374 - clear selection once a dataset has been removed

### DIFF
--- a/app/components/Dataset.tsx
+++ b/app/components/Dataset.tsx
@@ -56,22 +56,19 @@ export interface DatasetProps {
 const logo = require('../assets/qri-blob-logo-tiny.png') //eslint-disable-line
 
 class Dataset extends React.Component<DatasetProps> {
-  super () {
-    this.openWorkingDirectory = this.openWorkingDirectory.bind(this)
-    this.publishUnpublishDataset = this.publishUnpublishDataset.bind(this)
-    this.handleShowStatus = this.handleShowStatus.bind(this)
-    this.handleShowHistory = this.handleShowHistory.bind(this)
-    this.handleReload = this.handleReload.bind(this)
+  constructor (props: DatasetProps) {
+    super(props);
+
+    [
+      'openWorkingDirectory',
+      'publishUnpublishDataset',
+      'handleShowStatus',
+      'handleShowHistory',
+      'handleReload'
+    ].forEach((m) => { this[m] = this[m].bind(this) })
   }
 
   componentDidMount () {
-    // poll for status
-    this.statusInterval = setInterval(() => {
-      if (this.props.workingDataset.peername !== '' || this.props.workingDataset.name !== '') {
-        this.props.fetchWorkingStatus()
-      }
-    }, defaultPollInterval)
-
     // electron menu events
     ipcRenderer.on('show-status', this.handleShowStatus)
 
@@ -96,6 +93,12 @@ class Dataset extends React.Component<DatasetProps> {
 
     ipcRenderer.removeListener('reload', this.handleReload)
   }
+
+  statusInterval = setInterval(() => {
+    if (this.props.workingDataset.peername !== '' || this.props.workingDataset.name !== '') {
+      this.props.fetchWorkingStatus()
+    }
+  }, defaultPollInterval)
 
   handleShowStatus () {
     this.props.setActiveTab('status')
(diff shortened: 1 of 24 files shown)


app/components/Dataset.tsx
```diff
@@ -49,12 +49,21 @@ export interface DatasetProps {
...
   fetchWorkingStatus: () => Promise<ApiAction>
   fetchBody: (page: number) => Promise<ApiAction>
+  setRoute: (route: string) => Action
   fetchWorkingDataset: () => Promise<ApiAction>
 }
 
 const logo = require('../assets/qri-blob-logo-tiny.png') //eslint-disable-line
 
 class Dataset extends React.Component<DatasetProps> {
+  super () {
```
b5: interesting that we need to use `super` here.

test/app/utils/formValidation.test.ts
```diff
@@ -124,12 +125,16 @@ describe('formValidation', () => {
       err: ERR_INVALID_DATASETNAME_CHARACTERS
     },
     {
-      string: '👋🏻',
+      string: 'hi👋🏻',
```
b5: Interesting that this instance didn't fail. We should still fail on a string that starts with an emoji, but let's get that fixed later

ramfox: it did fail, it was just failing on an error about the string not starting with a letter, which we already test. I wanted it to fail on an error about unaccepted characters. I can add the other back in.

app/components/Dataset.tsx
ramfox: I pushed another commit to refactor to match our previous patterns for binding to `this`!

PR #378, opened 2019-12-23
feat(logging): Log to a file, menu item to export it

dustmop:
When launching the qri backend, always log its output to a temporary file. If a developer wants to see output in a terminal, they can either run `qri connect` themselves, or they can `tail -f` the log. The log filename uses the current day in the filename.

In development mode, try to locate an installed `qri` binary, which makes it easier to develop both desktop and the backend at the same time.

Add a menu item under 'Help' which lets the user copy the debug log to a path.

Output a few useful pieces of info: which version of the qri backend was launched, the path to the the backend, the path to the log file.

### DIFF
--- a/app/backend.js
+++ b/app/backend.js
@@ -10,42 +10,56 @@ class BackendProcess {
   constructor () {
     this.qriBinPath = null
     this.process = null
+    this.debugLogPath = null;
 
-    if (process.env.NODE_ENV === 'development') {
-      // if we're in dev mode write to this process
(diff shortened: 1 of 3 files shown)


dustmop:
Next step, add a flag to `qri connect` that turns on logging for everything, such as `qri connect --log-all` and add that to our backend invocation. Only way we can see what's going wrong all the time is by logging everything we can.

ramfox:
Fixed a dialog bug and rebased.

PR #380, opened 2020-01-03
test(e2e): expand and refine e2e tests

ramfox:
Outline of e2e flows: https://docs.google.com/document/d/1Ja_8br4I18K5qxBqI_qsDdkld8dK9xFBj6SVNc12Xio/edit#

This pr:
- gets the e2e tests to run in circleci! aka it will run in a docker image!
- ensures that the app has the correct id tags for the e2e tests to be able to hook into the dom
- does the first round of refactoring to the e2e tests to create helper functions that make the tests massively easier to understand
- cleans up the app and the tests to get the tests to pass!

Tests:
 - open window
- accept tos
- create new account
- create dataset from csv file

ps it is so cool knowing now that when tests pass on circle ci, it means that the e2e tests are passing!!!


test/e2e/e2e.spec.ts
```diff
@@ -35,201 +48,159 @@ describe('Qri End to End tests', function spec() {
...
+  // onTab checks to see if the element exists with the 'active' class
+  const onTab = async (tab: string) => {
+    expect(await app.client.element(`#${tab}-tab.active`)).toBeDefined()
+  }
+
+  // checkStatus ensures that component exists with the correct status dot class
+  const checkStatus = async (component: string, status: string) => {
+    expect(await app.client.element(`#${component}-status.status-dot-${status}`)).toBeDefined()
+  }
 })
```
b5: These should be moved into a separate file & destructured on import:

```js
import { onTab, checkStatus, ... } from './spectronTestUtils.ts`
```

test/e2e/e2e.spec.ts
```diff
@@ -4,23 +4,36 @@ import fs from 'fs'
 import TestBackendProcess from '../utils/testQriBackend'
 import fakeDialog from 'spectron-fake-dialog'
 
-const { Application } = require('spectron');
+const { Application } = require('spectron')
 
-const delay = (time: number) => new Promise(resolve => setTimeout(resolve, time))
+const delay = async (time: number) => new Promise(resolve => setTimeout(resolve, time))
```
b5: let's group this with the test utility functions you've defined

PR #390, opened 2020-01-09
feat(TypePicker): create TypePicker and other needed components

ramfox:
run `yarn storybook` to view component

added components:
- **overlay/Overlay**: used as a pop over, will call `onCancel` if you click outside fo the div, or click the 'x'. The `navigation` prop is optional. Any `children` will overflow-y scroll.
- **item/DataType** defines the `DataTypes` type. Wraps the type name, description, and icon. `type` and `description`, are defined and passed down. May want to refactor to keep description and type defined here, and just use the type to determine all content.
- **nav/TabPicker**: has size options 'sm' and 'md', and color options 'light' and 'dark'
- **overlay/TypePickerOverlay**: a version of `Overlay` that lists the DataTypes and passes through the `navigation` component. `typesAndDescriptions` are defined here.
- **structure/ColumnType**: the component used to show the column type with its icon. Is not 'clickable' if there is no 'onClick' func passed down. I a column has more than one type, it will indicate 'multi'
- **structure/TypePicker**: takes all these components and creates a TypePicker. Click on the columnType to get a TypePickerOverlay popup. Choose between single or multi mode. The chosen type is kept in the hidden input named for the column (`${column-name}-type`)

This commit also created scss files based on the folder structure:
- chrome
- item
- nav
- overlay
- structure
- type

It also adds a story for each component under the `structure` tab


app/components/structure/ColumnType.tsx
```diff
@@ -0,0 +1,36 @@
...
+const ColumnType: React.FunctionComponent<ColumnTypeProps> = ({
+  type = 'any',
+  onClick,
+  active = false
+}) => {
+  let shownType = ''
+  if (Array.isArray(type)) {
+    if (type.length > 1) {
+      shownType = 'multi'
+    } else {
```
dustmop: Should have a branch for `type.length == 0`, just to avoid the undefined case.

stories/3-Structure.stories.tsx
```diff
@@ -0,0 +1,110 @@
...
+    <div style={{ display: 'flex', flexDirection: 'column', justifyContent: 'space-around', alignItems: 'center', height: 100 }}>
+      <TabPicker tabs={['single', 'multi']} size= 'sm' color='dark' onClick={() => {}} activeTab='single'/>
+      <TabPicker tabs={['status', 'history']} size= 'md' color='light' onClick={() => {}} activeTab='status'/>
+    </div>
+  )
+}
+
+tabPicker.story = {
+  name: 'Tab Picker',
+  parameters: { note: 'used in picker pop-over, but also can be used whereever we need tabs' }
```
dustmop: Spelling: whereever -> wherever

stories/3-Structure.stories.tsx
```diff
@@ -0,0 +1,110 @@
...
+import TypePicker from '../app/components/structure/TypePicker'
+
+export default {
+  title: 'Structure',
+  parameters: {
+    notes: ''
+  }
+}
+
+const allColumnTypes: DataTypes[] = ['string', 'integer', 'number', 'boolean', 'null', 'any']
```
dustmop: No 'array' or 'object'? Could you add a comment mentioning why they're not here?

app/components/item/DataType.tsx
```diff
@@ -0,0 +1,28 @@
+import * as React from 'react'
+import Icon from '../chrome/Icon'
+
+export type DataTypes = 'string' | 'integer' | 'number' | 'boolean' | 'null' | 'any' | 'array' | 'object'
```
dustmop: This list of types appears in four places: here, app/components/overlay/TypePickerOverlay.tsx, app/components/structure/TypePicker.tsx, and stories/3-Structure.stories.tsx. Could we line-up them up everywhere so that they always follow the same order? It seems like app/components/structure/TypePicker.tsx has the order that matters most, which would mean changing it here.

app/components/overlay/Overlay.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+      if (!isIn) {
+        e.stopPropagation()
+        onCancel()
+      }
+    }
+
+    window.addEventListener('click', isInOverlay)
+    return (
+      window.addEventListener('click', isInOverlay)
```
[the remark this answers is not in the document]
ramfox: yes thank you... how... was this working?

stories/3-Structure.stories.tsx
ramfox: oversight, just kept forgetting them. adding them in.

app/components/item/DataType.tsx
ramfox: Simplified this. There is one place defining the different strings that count as `DataTypes`, that is in the `DataType` component referenced here.

For the actual list of types (defining the order and also adding descriptions), we reference the `typesAndDescriptions` array defined in `TypePicker`

PR #394, opened 2020-01-15
fix(commit, MetadataEditor): fix meta editing and the flow after you commit

ramfox:
Adjust the `saveWorkingDatasetAndFetch` and `Commit` components, along with the `mutations` reducer and actions to fix the flow after you commit.

We need to wait for both the commit to be successful and for the history list to update, inorder to show the user a clean commit experience. To do that, we've added a `saveComplete`/`SAVE_COMPLETE` action to the `mutations` reducer that gets called when save & history fetching is complete

Also adjusts the MetadataEditor to fix a content jumping problem

Also, also fixes a bug when navigating to a dataset using the dataset list.

After this pr, e2e tests are passing locally, but not on ci

### DIFF
--- a/app/actions/api.ts
+++ b/app/actions/api.ts
@@ -6,6 +6,7 @@ import { CALL_API, ApiAction, ApiActionThunk, chainSuccess } from '../store/api'
 import { SelectedComponent, MyDatasets } from '../models/store'
 import { actionWithPagination } from '../utils/pagination'
 import { openToast, setImportFileDetails } from './ui'
+import { setSaveComplete } from './mutations'
 import { setWorkingDataset, setSelectedListItem, setActiveTab, clearSelection } from './selections'
 import {
   mapDataset,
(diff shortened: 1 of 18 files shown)


app/actions/api.ts
```diff
@@ -443,18 +444,23 @@ export function saveWorkingDatasetAndFetch (): ApiActionThunk {
...
+      dispatch(setSaveComplete(action.payload.err.message))
       dispatch(openToast('error', 'commit', action.payload.err.message))
       throw action
     }
+    if (getActionType(response) === 'success') {
+      dispatch(setSaveComplete())
+      dispatch(openToast('success', 'commit', 'commit success!'))
+      dispatch(setSelectedListItem('commit', path))
+      dispatch(setActiveTab('history'))
+      dispatch(setSelectedListItem('commitComponent', DEFAULT_SELECTED_COMPONENT))
```
b5: hmmm... these are all firing in parallel? Let's examine this in a future PR!

test/e2e/e2e.spec.ts
```diff
@@ -206,32 +206,59 @@ describe('Qri End to End tests', function spec () {
     await checkStatus('structure', 'added')
   })
 
+  it ('navigate to collection and back to dataset', async () => {
```
b5: 🙌

PR #402, opened 2020-01-22
refactor(Dataset): move fetching into Dataset and remove Containers

### LINKED ISSUE #400: Bug: more routing/fetching issues
The way we are using the router + selections is sort of muddled right now. Eventually I think the way forward is to rely on the router and remove the selection reducer, but for now this selections tree is being relied on most places.

When a dataset returns with `repo: no history`, we need to display the `no history` page

When we click on a dataset from the collections page, we aren't triggering the dataset to fetch from head.

When we create a new dataset, we aren't triggering fetching.

Everything displayed on the commit details page should come from the commit details state tree, let's stop mixing selections with this.

Refactor so DatasetContainer is the only place we are pulling from the state tree in the Dataset Component, and the Dataset component is the only place we are fetching.

```typescript
// Dataset props:
Dataset: {
  data: {
    workingDataset: (workingDataset),
    versionedDataset: (commitDataset),
    history: (workingDataset.history)
  },
  // display details, eg sidebarWdith
  // fetching functions, eg fetchWorkingDataset
  ...
}
```

In Dataset, `ComponentDidUpdate`:

```typescript
// if selections have changed, let's fetch all the things we can immediately
if (selections.peername !== workingDataset.peername || selections.name !== workingDataset.name) {
  fetchingWorkingDataset
  fetchHistory
  fetchStats
  fetchWorkingStatus
}

// if selections have changed, including the commit, fetch VersionedDataset
if (selections.peername !== versionedDataset.peername || selections.name !== versionedDataset.name || selections.commit !== versionedDataset.path) {
  fetchVersionedDataset
}

// if history has changed, set the commit in the selections
if (prevHistory.value.length === 0 && history.value.length !== 0) {
  setCommit(history.value.length[0])
}

// in fetchVersionedDataset, if there is no commit available or selected, just zero out the version, don't fetch
in fetchVersionedDataset: 
{
  ...
  if (selections.commit === '') {
    dispatch({type: CLEAR_VERSIONED_DATASET, peername: selections.peername, name: selections.name})
  }
  // otherwise fetch as normal
}

// commitDetails reducer
  case CLEAR_VERSIONED_DATASET:
    return {
      ...initialState,
      peername: action.peername,
      name: action.name,
    }
```

* [x] Dataset: adjust props in component and container, including passing down more fetching functions
* [x] Dataset: add all fetching into `ComponentDidUpdate`
* [x] DatasetSidebar: adjust props in DatasetSidebar, remove any unneeded fetching, and in Dataset, pass props directly rather then through DatasetSidebarContainer
* [x] CommitDetails: adjust props in component & remove any unneeded fetching, in Dataset: pass props directly through rather then using CommitDetailsContainer
* [x] Audit to make sure we don't have any unneeded calls to any of our fetching actions that change workingDataset or commitDataset unexpectedly

ramfox:
- refactor props to `Dataset`,` DatasetSidebar`, `CommitDetails` to move to `data` paradigm
- remove `DatasetSidebarContainer`, `CommitDetailsContainer`
- add fetching to `Dataset`  in `ComponentWillMount` and `ComponentDidMount` funcs, and remove fetching from `App` and `bootstrap`
- adjust `Commit` fetching functions to clear the `CommitDetails` if there is no commit selected
- add `setCommit` function that wraps `setSelectedListItem`
- remove fetching from `fsiWrite`, it's unnecessary
 - change `isLoading` in `CommitDetails` `initialState` to false, otherwise the fetching doesn't go correctly
- remove `commit` from localStorage, going to slowly remove these as it comes up
- added e2e test for testing fetching

closes #400


app/reducers/commitDetail.ts
```diff
@@ -43,6 +43,8 @@ const [COMMITSTATUS_REQ, COMMITSTATUS_SUCC, COMMITSTATUS_FAIL] = apiActionTypes(
 const [COMMITBODY_REQ, COMMITBODY_SUCC, COMMITBODY_FAIL] = apiActionTypes('commitbody')
 const [COMMITSTATS_REQ, COMMITSTATS_SUCC, COMMITSTATS_FAIL] = apiActionTypes('commitstats')
 
+export const CLEAR_VERSIONED_DATASET = 'CLEAR_VERSIONED_DATASET'
```
b5: Can this be `CLEAR_DATASET_HEAD` instead?

app/reducers/selections.ts
```diff
@@ -18,7 +18,7 @@ export const initialState: Selections = {
   name: localStore().getItem('name') || '',
   activeTab: localStore().getItem('activeTab') || 'status',
   component: localStore().getItem('component') as SelectedComponent || '',
-  commit: localStore().getItem('commit') || '',
+  commit: '',
```
b5: This change to me represents a vital one. Let's talk about rethinking the way local storage works entirely in a future PR. I think it should load at the beginning and end of a session

PR #405, opened 2020-01-23
fix(body): refine props to Body components, fix fetching status

ramfox:
- `fetchingWorkingStatus` relies on `workingDataset` for `fsiPath` before it can fetch. This should be refactored/detangled in a future pr, for now, wait for workingDataset to return before fetching status
- before `status` returns, display all the components (disabled) in the status pane
- Refine the props passed down to Body, ComponentList, and remove the BodyContainer
- fix bug that had CommitDetails reducer write over the `body` component when the `DATASET_SUCC` action returned, zeroing out the body, even if it was correct


app/components/CommitDetails.tsx
```diff
@@ -98,7 +107,19 @@ const CommitDetails: React.FunctionComponent<CommitDetailsProps> = ({
         )}
         sidebarWidth={150}
         mainContent={(
-          <DatasetComponent isLoading={loading} component={selectedComponent} componentStatus={status[selectedComponent]} history />
+          <DatasetComponent
+            data={data}
+
```
b5: can we nix these blank lines?

app/components/DatasetComponent.tsx
```diff
@@ -16,9 +15,24 @@ import CommitContainer from '../containers/CommitContainer'
...
+import { Action } from 'redux'
 
 interface DatasetComponentProps {
+  data: CommitDetails
+
+  // display details
+  details: Details
+
+  // seting actions
+  setDetailsBar: (details: {[key: string]: any}) => Action
```
b5: Use a [Record](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkt) instead of object with string keys.

```suggestion
  setDetailsBar: (details: Record<string,any>) => Action
```

app/components/Body.tsx
```diff
@@ -56,26 +46,22 @@ const extractColumnHeaders = (structure: any, value: any): undefined | object =>
 
   return schema &&
     schema.items &&
-    schema.items.items.map((d: { title: string }): string => d)
+    schema.items.items.map((d: { title: string }): {[key: string]: string} => d)
```
b5: ditto for the `Record` comment

app/components/CommitDetails.tsx
```diff
@@ -44,18 +46,25 @@ const CommitDetailsHeader: React.FunctionComponent<CommitDetailsHeaderProps> = (
 export interface CommitDetailsProps {
   data: ICommitDetails
   selections: Selections
+  details: Details
 
+  fetchCommitBody: (page?: number, pageSize?: number) => ApiActionThunk
+  setDetailsBar: (details: {[key: string]: any}) => Action
```
b5: `Record<string,any>` here as well

app/components/Dataset.tsx
```diff
@@ -56,6 +58,7 @@ export interface DatasetProps {
   setRoute: (route: string) => Action
   setCommit: (path: string) => Action
   setComponent: (type: ComponentType, activeComponent: string) => Action
+  setDetailsBar: (details: {[key: string]: any}) => Action
```
b5: `Record`

app/components/Dataset.tsx
```diff
@@ -123,10 +126,17 @@ class Dataset extends React.Component<DatasetProps> {
...
-      fetchWorkingDataset()
+      // TODO (ramfox): refactor action so that `fetchWorkingStatus` does not rely on workingDataset
+      new Promise((resolve, reject) => {
+        try {
+          fetchWorkingDataset()
+        } catch (e) {
+          reject(e)
+        }
+        resolve()
+      }).then(() => fetchWorkingStatus())
```
b5: Instead of injecting the same code in two places, create another action. How about:

```typescript
async function fetchWorkingDatasetAndStatus() {
  return fetchWorkingDataset().then(fetchWorkingStatus)
}
```

I think the `promise/try/catch` combo you're using here is effectively a noop. Should be able to re-write to just `fetchWorkingDataset.then(fetchWorkingStatus)`. Maybe I'm missing something.

PR #428, opened 2020-02-03
refactor(routes): refactor top-level routes

b5:
working on better route representation, moving each route into it's respective place. The biggest change here is renaming "/datset" to "/workbench", and moving all onboarding stuff into it's own "/onboarding" prefixed routes


app/components/workbench/Workbench.tsx
```diff
@@ -1,48 +1,58 @@
...
+
+export interface WorkbenchData {
+  // TODO (b5) - this is a shim to get selection data out of the workbench component
+  // these should be factored into the given workingDataset
+  location: string
+  tab: 'status' | 'history'
+  peername: string
+  name: string
+  path: string
+  component: string
```
ramfox: Is this just an example of what we want to happen in the future? because none of these props are actually being used in the component. Right now all of that info is being pulled directly from the `selections` props that is being passed from the container.

app/components/workbench/Workbench.tsx
```diff
@@ -509,4 +448,4 @@ class Dataset extends React.Component<DatasetProps> {
   }
 }
 
-export default withRouter(Dataset)
+export default withRouter(Workbench)
```
ramfox: we shouldn't need `withRouter` anymore, since we wrap the `WorkbenchContainer` in a `Route`, the routing props should come from there.

app/containers/WorkbenchContainer.tsx
```diff
@@ -50,7 +52,16 @@ const DatasetContainer = connect(
...
+    const data: WorkbenchData = {
+      // TODO (b5) - here we're intentionally constructing location from selection state,
+      // as it's more accurate than values derived from the route. Refactor away
+      // from selection toward using route match params
+      location: locationFromSelection(selections),
+      tab: props.tab || 'status',
+      peername: props.peername || selections.peername,
+      name: props.name || selections.name,
+      path: props.path || selections.commit,
+
```
ramfox: Noted elsewhere, but although this is being set here, it is not being used in the component.

app/routes.tsx
```diff
@@ -20,56 +20,91 @@ export default function Routes (props: any) {
...
-          <Route exact path='/collection' render={() => {
-            ipcRenderer.send('show-dataset-menu', false)
-            if (!hasAcceptedTOS) return <Redirect to='/' />
-            if (!qriCloudAuthenticated) return <Redirect to='/signup' />
-            return <CollectionContainer setModal={setModal} />
-          }}/>
+        <Route exact path='/workbench' render={() => {
+          return sectionElement('workbench', <WorkbenchContainer
+            tab="status"
+            setModal={setModal}
```
ramfox: `setModal` is not being used in the container, we end up importing it.

app/routes.tsx
```diff
@@ -20,56 +20,91 @@ export default function Routes (props: any) {
...
+        <Route path='/workbench/:username/:dataset' render={(props) => {
+          const { params } = props.match
+          return sectionElement('workbench', <WorkbenchContainer
+            location={props.match.path}
+            tab="status"
+            username={params.username}
+            dataset={params.dataset}
+            path={params.path}
+            component={params.component}
+            setModal={setModal}
```
ramfox: personally think we should just pass {...props} down to `WorkbenchContainer` (whose props are now `RouteComponetProps` which gives us `location` & `match`, and do all this logic down in WorkbenchContainer. Feels likes to understand what's going on we have to look at multiple places.

test/e2e/e2e.spec.ts
```diff
@@ -9,6 +11,12 @@ const { Application } = require('spectron')
 describe('Qri End to End tests', function spec () {
   let app: any
   let backend: any
+  let registry: any
+  const imagesDir = process.env.E2E_ARTIFACTS_PATH || os.tmpdir()
```
ramfox: doooope

test/utils/e2eTestUtils.ts
```diff
@@ -41,13 +42,23 @@ const delay = async (time: number) => new Promise(resolve => setTimeout(resolve,
...
+      }, 10000, `expected url to be '${expected}', got: ${location.hash || location}`)
+    } catch (e) {
+      if (screenshotLocation) {
+        app.browserWindow.capturePage().then(function (imageBuffer) {
+          console.log(`writing screenshot: ${screenshotLocation}`)
+          fs.writeFileSync(screenshotLocation, imageBuffer)
+        })
+      }
+      throw e
+    }
```
ramfox: dooooope

test/utils/testTempRegistry.ts
```diff
@@ -0,0 +1,101 @@
...
+  constructor () {
+    this.registryBinPath = ''
+    this.process = null
+    this.dir = ''
+  }
+
+  start () {
+    // const { resourcesPath = '' } = process
+    // Locate the binary for the qri backend command-line
+    // this.registryBinPath = this.findTempRegistryBin(['qri', resourcesPath, path.join(__dirname, '../../')])
```
ramfox: remove or explain why commented out plz

test/utils/testTempRegistry.ts
```diff
@@ -0,0 +1,101 @@
...
+    this.registryBinPath = this.findTempRegistryBin()
+    if (this.registryBinPath === '') {
+      throw `couldn't find a temp_registry_server binary on your $PATH, please install qri backend to run e2e tests`
```
ramfox: Is this explanation up to date? Do you just need the qri backend for this to run? Thought it was a separate binary.

test/utils/testTempRegistry.ts
```diff
@@ -0,0 +1,101 @@
...
+    // this.teardownRepo()
+    if (this.process) {
+      this.process.kill()
+      this.process = null
+    }
+  }
+
+  launchProcess () {
+    try {
+      // const [ base, qriPath, ipfsPath ] = this.setupRepo()
```
ramfox: remove or explain why commented out plz

test/utils/testTempRegistry.ts
```diff
@@ -0,0 +1,101 @@
...
+      this.logPath = path.join(this.dir, 'temp_registry_server.log')
+      const write = fs.createWriteStream(this.logPath)
+      this.stdout = write
+      this.stderr = write
+      this.process = childProcess.spawn(this.registryBinPath, ['connect', '--setup'], {
+        // stdio: ['pipe', this.stdout, this.stderr],
```
ramfox: remove or explain why commented out plz

test/utils/testTempRegistry.ts
```diff
@@ -0,0 +1,101 @@
...
+      this.process.stdout.pipe(this.stdout);
+      this.process.stderr.pipe(this.stderr);
+
+      this.process.on('error', (err: any) => { this.handleEvent('error', err) })
+      this.process.on('exit', () => { this.handleEvent('exit', 'registry exited') })
+      this.process.on('close', (err: any) => {
+        this.handleEvent('close', err)
+        write.close()
+        const buffer = fs.readFileSync(this.logPath)
+        console.log(buffer.toString())
```
ramfox: is this what you were mentioning, that the registry log gets dumped to console? If we have the `temp_registry_server.log` file, then I agree we don't need this.

app/components/Transform.tsx
```diff
@@ -16,7 +16,7 @@ export interface TransformProps {
 }
 
 const Transform: React.FunctionComponent<TransformProps> = (props) => {
-  const { value } = props
+  const { value = '' } = props
```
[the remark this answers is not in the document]
b5: it _shouldn't_ be optional, but something is in fact passing a null value. I'll make a note.

test/utils/testTempRegistry.ts
b5: lol oops, forgot to read the second part of that sentence

PR #434, opened 2020-02-06
feat(QriRef): rename DetailedDatasetRef to VersionInfo, introduce QriRef

b5:
We've been working on the refString -> QriRef -> VersionInfo -> Dataset chain of data enrichment.

This commit is a first-cut to introduce two data models we intend to rely on heavily as we tie more things back to the router

### DIFF
--- a/app/actions/mappingFuncs.ts
+++ b/app/actions/mappingFuncs.ts
@@ -17,26 +17,28 @@ export function mapVersionInfo (data: any[]): VersionInfo[] {
   return data.map((ref: any) => ref as VersionInfo)
 }
 
-// detailedDatasetRefToDataset converts a detailed ref to a sparsely-populated
+// versionInfoToDataset converts a detailed ref to a sparsely-populated
 // dataset object without using any fetching
-export function detailedDatasetRefToDataset (ddr: VersionInfo): Dataset {
+export function versionInfoToDataset (vi: VersionInfo): Dataset {
+  const theme = (vi.themeList) ? vi.themeList.split(',', -1) : undefined
+
   return {
-    username: ddr.username,
-    name: ddr.name,
-    path: ddr.path,
+    username: vi.username,
+    name: vi.name,
+    path: vi.path,
     meta: {
-      title: ddr.metaTitle,
-      theme: ddr.themeList && ddr.themeList.split(',', -1)
+      title: vi.metaTitle,
+      theme
     },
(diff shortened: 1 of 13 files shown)


app/actions/mappingFuncs.ts
```diff
@@ -13,13 +13,13 @@ export function mapRecord (data: any): Record<string, string> {
...
-export function mapDetailedDatasetRef (data: any[]): DetailedDatasetRef[] {
-  return data.map((ref: any) => ref as DetailedDatasetRef)
+export function mapVersionInfo (data: any[]): VersionInfo[] {
+  return data.map((ref: any) => ref as VersionInfo)
 }
 
 // detailedDatasetRefToDataset converts a detailed ref to a sparsely-populated
 // dataset object without using any fetching
-export function detailedDatasetRefToDataset (ddr: DetailedDatasetRef): Dataset {
+export function detailedDatasetRefToDataset (ddr: VersionInfo): Dataset {
```
ramfox: rename to `versionInfoToDataset`

app/models/qriRef.ts
```diff
@@ -0,0 +1,63 @@
+import { RouteComponentProps } from "react-router"
+
+// QriRef models a reference to a user, dataset, or part of a dataset within Qri
+// We use "QriRef" instead of "Ref" to disambiguate with the react "ref"
+// property.
+//
+// QriRef are structured expansions of a reference string, breaking elements of
+// a reference into fields. For example the folling reference string:
```
ramfox: "folling" => "following"

app/components/network/NetworkHome.tsx
```diff
@@ -45,21 +49,21 @@ const NetworkHome: React.FC<RouteComponentProps> = ({ history }) => {
...
-          history.push(`/network/${peername}/${name}`)
+          push(`/network/${peername}/${name}`)
         }} data={ds} key={i} />)}
       </div>}
     </div>
   )
 }
 
-export default NetworkHome
+export default withRouter(NetworkHome)
```
ramfox: just checking that this is the pattern we want. You get `QriRef` s from your container, but `push` functions from `withRouter`?

b5: still playing with that one, and the answer is I'm not totally sure. I'm working on that now, will follow up with a PR that we can discuss.

PR #466, opened 2020-03-02
style: fix search modal overflow bug

chriswhong:
Prevents a long list from overflowing the modal.


b5:
ah shit, that's wrong. probably should have been `fix(SearchModal)`. my bad.

PR #469, opened 2020-03-03
fix(workbench): viewing transform component shouldn't crash 

b5:
I ended up doing a bunch of refactoring at the same time, mainly aimed at merging container definitions. Damn that DatasetComponent needs to drop in size!


app/components/Navbar.tsx
```diff
@@ -108,15 +114,6 @@ const Navbar: React.FunctionComponent<NavbarProps> = (props: NavbarProps) => {
     }
   ]
 
-  if (__BUILD__.ENABLE_NETWORK_SECTION) {
```
ramfox: oof yep let's get rid of this

app/reducers/workingDataset.ts
```diff
@@ -102,7 +102,7 @@ const workingDatasetsReducer: Reducer = (state = initialState, action: AnyAction
             value: dataset && dataset.structure ? dataset.structure : undefined
           },
           transform: {
-            value: dataset && dataset.transform ? dataset.transform : undefined
+            value: dataset && dataset.transform ? { scriptBytes: atob(dataset.transform.scriptBytes) } : undefined
```
ramfox: I'm a bit thrown off by the `scriptBytes` field potentially having different meanings depending on where it is in our stack.

In backend world it's a slice of bytes
In an http response, it's a base64 encoded string
In frontend it's an un-encoded string

For readme, we are sidestepping this issue by setting the entire value of the `readme` component to its string contents. However, `transform` potentially more details (its format), so I understand why it's not following the same pattern.

However, I find it confusing and wanted to note that here.

Another option is to potentially not unencode the `scriptBytes` field until we get down to the component that actually displays the content (for `readme` and `transform`), rather than doing it on ingest in our state tree. Then there are two `scriptBytes` states we know about, in backend it is a byte slice and in frontend it's a base 64 encoded string. 

Curious about your thoughts.

app/routes.tsx
```diff
@@ -54,21 +54,16 @@ export default function Routes (props: any) {
         }} />
 
         {/* App Sections */}
-        { __BUILD__.ENABLE_NETWORK_SECTION &&
```
ramfox: ty

app/reducers/workingDataset.ts
ramfox: Also this needs to be added to the `CommitDetails` reducer as well, or we won't show transforms in the history tab!

ramfox: Oh wait also on further review, we aren't keeping `format`. Definitely believe whatever we do with one component we do with the other, then.

ramfox: Or is it semantically the same to have scriptBytes be a string in frontend because in go a slice of bytes and a string are so similar, and this shouldn't be ringing any bells in my head? Having so many thoughts about this.

b5: Great catch on the `commitDetails` reducer! For now I think we should just keep the pattern you've been using for `readme`. I tried to get fancy. Don't think it's worth it 😄

b5:
Ok dope merging

PR #470, opened 2020-03-04
feat(compare): compare two csv files with compare section

b5:
builds on #469. I'll rebase after it's merged

Things to check & fix:
* [ ] centering / layout for diff display table
* [ ] should show "these files are equal" when diff script contains no changes. Can look at diff stat for this

### DIFF
--- a/app/components/compare/CompareSidebar.tsx
+++ b/app/components/compare/CompareSidebar.tsx
@@ -1,0 +1,101 @@
+import React from 'react'
+import { remote } from 'electron'
+
+import TextInput from '../form/TextInput'
+import ButtonInput from '../form/ButtonInput'
+import Icon from '../chrome/Icon'
+
+export interface CompareParams {
+  left: string
+  right: string
+}
+
+export interface CompareSidebarProps {
+  data: CompareParams
+  onChange: (p: CompareParams) => void
+}
+
+const CompareSidebar: React.FC<CompareSidebarProps> = ({ data, onChange }) => {
+  const pathPicker = (side: string) => {
+    const window = remote.getCurrentWindow()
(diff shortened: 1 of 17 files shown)


app/components/compare/CompareSidebar.tsx
```diff
@@ -0,0 +1,136 @@
...
+      </div>
+    </div>
+  )
+}
+
+const mapStateToProps = (state: Store, ownProps: CompareSidebarProps) => {
+  return ownProps
+}
+
+// TODO (b5) - this component doesn't need to be a container. Just feed it the right data
```
b5: Stray comment

app/components/compare/TableDiff.tsx
```diff
@@ -1,107 +1,111 @@
...
-              if (open) {
+      <div id='table-diff-container'>
+        <table>
+          {data.schema && <TableDiffSchemaHead data={schema} />}
+          <tbody>
+            {data.diff.map((p, i) => {
+              if (Array.isArray(p[0])) {
+                count = 1
+                // leftCount = p[0][0][0]
+                // rightCount = p[0][1][0]
```
b5: Stray Comments

app/components/compare/TableDiff.tsx
```diff
@@ -1,107 +1,111 @@
-import React, { useState } from 'react'
+import React from 'react'
+
+import TableDiffSchemaHead from './TableDiffSchemaHead'
+import DiffStat from './DiffStat'
 
 interface TableDiffProps {
   data: any
```
b5: Ideally this would be a type definition for a diff script. One day...

ramfox: yez

PR #487, opened 2020-03-23
fix(MultiStructuredInput): edits to stop UI flickering when saving

### LINKED ISSUE #485: fix(MultiStructuredInput): bugs around blur, adding & removing row
If we send over an object where `meta.contributors` is `[{"id":"","name":"","email":""}]`, currently, backend translates this into empty object. If we fetch that dataset, the `meta.contributors` is `[{}]`. One could argue this should be handled by the backend, but for UI purposes on frontend, we should be able to handle the empty object as input regardless.

We are having issues adding and removing rows. When you add a second row, the row blips out of existence immediately. If the row add is successful, adding input to the fields & then blurring can sometimes cause the whole row to blip out as well. I haven't confirmed, but it feels as though this bug is a mix between how we are handling state (in `MultiStructuredInput`), as opposed to all other cases the lowest input is handling its own state. 

We currently use normal inputs in the `MultiStructuredInputRow`, and we keep state of the entire contributors/citations list in it's parent `MultiStructuredInput`. I believe we need to refactor to be closer to the model/relationship we have in `Schema`, `SchemaItem`, and `DynamicEditField`.

1) DynamicEditField controls its own state. We only fire off write calls on blur ('Enter' and 'Tab' keystrokes cause blur)
2) SchemaItem passes down a `handleDynamicEdit` function that integrates a specific input value into the larger row & calls a `write` function, making sure not to mutate state
3)  Schema passes down a `write` handler functions that integrates a specific row into the larger schema value.

To translate this to `MultiStructuredInput`, `MultiStructuredInputRow`, these are the changes that need to be made:
* [x]  `MultiStructuredInputRow` renamed to `MultiStructuredInputItem` and moved into the `items` folder
* [x] in `MultiStructuredInputItem`, replace inputs with `DynamicEditField`
* [x] in `MultiStructuredInputItem`, `handleWrite` should integrate a specific field into the larger row. Pass this down to `DynamicEditField`'s `onChange` func
* [x] in `MultiStructuredInputItem`, when we get an empty object, or an object that does not contain all fields, still display each field
* [x] in `MultiStructuredInput`, remove state, pass down items directly
* [x] in `MultiStructuredInput`, pass down `onRemove` function to `MultiStructuredInputItem` that wraps the `write` function which slices out the target row (making sure not to mutate state)
* [x] in `MultiStructuredInput`, `write` function. If a row is empty (each of it's fields are the empty string, or the object is empty), remove it before writing (make sure the row exists before removing, in the case of adding random rows). 
* [x] in `MultiStructuredInput`, only allow users to add 1 empty row. Only state we need, keep track of the number of rows that we are displaying: `numRows`. If the number of rows we get from the state tree is equal to `numRows`, we can allow users to add a row, if not, disable adding a row
* [x] in `MultiStructuredInput`, `onAdd` just adds one to `numRows`, `numRows` is how we determine how many `MultiStructuredInputItem`s to display, any row that doesn't have a corresponding data item gets passed in the empty object, which `MultiStructuredInputItem` knows how to handle.

ramfox:
## edit:
this pr has been subsumed by #491 
closing it as it is no longer needed

----------

This pr does three main things:

1) tests each meta input for fsi and non-checked out datasets
2) tests the csv structure (format config + schema) for fsi and non-checked out datasets
3) pulls e2e testing sequences into reusable functions
  - `editCommit` - that checks the component status, writes the commit title and message, commits, and checks the title and component status are correct in history
  - `editMeta` - switches to the meta component, adds input for each meta field that the passed in dataset has filled, commits (using `editCommit`) and checked each field in history for accuracy
  - `editStructure` - switches the structure component, inputs each structure fields that the passed in dataset has filled (ensure the correct format config options are checked, fills out the schema), commits, and checks each editing field in history for accuracy

This pr also closes #485 - fixes UI flickering when `fsiWrite` happens. Also, renames `MultiStructuredInputRow` to `MultiStructuredInputItem`, and reactors `MultiStructuredInputItem` and `MultiStructuredInput` to use tables rather then flex. Also replaces the inputs in `MultiStructuredInputItem` to use our `DynamicEditField` components


b5:
@Kasey any clues on why CI keeps failing? Flakiness?

PR #491, opened 2020-03-24
fix(websockets): use websocket responses more intelligently

### LINKED ISSUE #485: fix(MultiStructuredInput): bugs around blur, adding & removing row

ramfox:
Welcome to this giant pr that fixes some stuff and tests a ton.

### e2e tests
This pr adds tests for meta & structure. It adds `editCommit`, `editMeta`, and `editStructure`, our first iteration on trying to breakup the e2e tests for reusability

`editCommit` - that checks the component status, writes the commit title and message, commits, and checks the title and component status are correct in history
`editMeta` - switches to the meta component, adds input for each meta field that the passed in dataset has filled, commits (using editCommit) and checked each field in history for accuracy
`editStructure` - switches the structure component, inputs each structure fields that the passed in dataset has filled (ensure the correct format config options are checked, fills out the schema), commits, and checks each editing field in history for accuracy

### WebSockets & fetching

While the websocket interface between the desktop and the backend is in its infancy, and so we expect the api to evolve and become more robust, it doesn't mean we shouldn't use the tools we have currently to minimize the amount of refetching the desktop needs to be doing.

Also, the method for editing in-app is much more reliable then using fsi, and they are too divergent to reason about well at a glance. Let's take the things that are reliable about in app editing and apply them to fsi editing, and use the websocket messages in a more logic way so we can minimize refetching and have the whole editing/saving process be more stable.

To that effect: this pr has in-app editing and fsi editing be almost identical (aka using the mutations state tree to keep track of changes and status) except for a few things:
- if we are checked out, along with each `mutations` edit, we also call `fsiWrite` to keep the underlying file updated
- if we are in the process of `writing` to the underlying file or we are `saving` (making a commit), ignore any websocket messages about that dataset while the write or commit is occuring. Basically, while we are in app, favor the edits that are being made in app, or if we are in the process of commiting, ignore any modified messages as these are most likely changes that are occuring during the save process.
- if we are NOT in the process of `writing` to the underlying file or commiting our changes, and we get a websocket message saying some component has been modified, created, or removed, and it is the dataset we are currently working on, we reset the mutations dataset and mutations status & re-fetch the underlying file. (This is the initial intended use for the websocket communication). This occurs in wsMiddleware and so has no effect on in-app editing.
- if a user needs to discard changes, discard the relevant mutations dataset and status sections, adding a call to the `fsiDiscardChanges` if we are checked out. The change in the underlying file will trigger a re-fetch to get us in the correct state, again in the wsMiddleware, and so has no effect on in-app editing.

This gets us a more reliable editing experience and fewer fetches!

### refactors

- Removes `fetchWorkingDatasetAndStatus` - Splits `fetchWorkingStatus` from `etchWorkingDataset`. These were stacked because `fetchWorkingStatus` needs to know the `fsi` state in order to fetch correctly. However, in other fetch calls, we are using the dataset list to know if we have `fsi` true or false. While I don't think this is the best metric, I'd rather have each function be consistent when we do get to a refactor and/or we don't need to pass an `fsi` param to the backend

- add `isWriting` flag to workingDataset, to track if we are writing to the filesystem. If we are writing, ignore any ws modified, created, or removed messages about this dataset.

- add `isSaving` flag to workingDataset to track if we are saving. If we are saving, ignore any ws modified, created, or removed messages about this dataset.

- fixes some typos in e2e tests

### fixes

This pr also closes #485 - fixes UI flickering when fsiWrite happens. Also, renames MultiStructuredInputRow to MultiStructuredInputItem, and reactors MultiStructuredInputItem and MultiStructuredInput to use tables rather then flex. Also replaces the inputs in MultiStructuredInputItem to use our DynamicEditField components

- fixes bug after checkout that doesn't refetch the dataset list before re-fetching dataset (giving us the incorrect fsi status)

- fixes isLoading bug in DatasetComponent

- fixes api change in history list and the `LogList` fetch response, now uses a mapHistory function to map the response to the standard `VersionInfo`


app/actions/api.ts
```diff
@@ -303,8 +293,18 @@ export function fetchHistory (page: number = 1, pageSize: number = pageSizeDefau
...
 export function fetchWorkingStatus (): ApiActionThunk {
   return async (dispatch, getState) => {
-    const { workingDataset } = getState()
-    const { peername, name, fsiPath } = workingDataset
+    const { selections, myDatasets } = getState()
+    const { peername, name } = selections
+
+    if (peername === '' || name === '') {
+      return Promise.reject(new Error('no peername or name selected'))
+    }
```
b5: In a future PR I'd love to factor selections out & pass a selection to `fetchWorkingStatus` as an argument instead. This pattern of rejecting based on internal state would be less magical.

app/actions/api.ts
```diff
@@ -303,8 +293,18 @@ export function fetchHistory (page: number = 1, pageSize: number = pageSizeDefau
...
+    // look up the peername + name in myDatasets to determine whether it is FSI linked
+    const dataset = myDatasets.value.find((d) => (d.username === peername) && (d.name === name))
```
b5: I'd love to see this as a selector that just takes state & arguments:
```typescript
const SelectCollectionDataset = function (state: State, peername: string, name: string): Dataset | undefined {
  return state.myDatasets.value.find((d) => (d.username === peername) && (d.name === name))
}
```

Not sure where that selector belongs tho. We could tackle this as part of a future PR

app/actions/mappingFuncs.ts
```diff
@@ -18,6 +18,19 @@ export function mapVersionInfo (data: any[]): VersionInfo[] {
...
+    const vinfo = {
+      ...item,
+      ...item.versionInfo
+    }
+    if (item.versionInfo) {
+      delete vinfo.versionInfo
+    }
+    return vinfo
+  })
+}
```
b5: We've chatted on slack a bunch about how this came as the cause of an unknown API break on the backend. In situations where you're writing code to make something work because of confusion elsewhere, I'd love a big `TODO (ramfox)` comment talking about why this function was created.

app/components/form/MultiStructuredInput.tsx
```diff
@@ -12,7 +14,7 @@ export interface MultiStructuredInputProps {
   labelTooltip?: string
   name: 'citations' | 'contributors'
   value: User[] | Citation[] | undefined
```
b5: interesting that this component is called `MultiSturcturedInput`, which implies it's a generic input component, but only works with `User` and `Citation` types.

Based on that I'd recommend renaming the component, what about `MetadataMultiInput`?

app/components/workbench/Workbench.tsx
```diff
@@ -148,58 +148,10 @@ class Workbench extends React.Component<WorkbenchProps, Status> {
...
-
-      // reset components
-      if (componentsToReset.includes('body')) fetchBody(-1)
-      if (
-        componentsToReset.includes('structure') ||
-        componentsToReset.includes('meta') ||
-        componentsToReset.includes('readme') ||
-        componentsToReset.includes('transform')
-      ) fetchWorkingDataset()
-    }
```
b5: wow, nice pickup!

app/actions/api.ts
ramfox: yes love it.

app/actions/mappingFuncs.ts
ramfox: done and done

app/components/form/MultiStructuredInput.tsx
ramfox: had the same thought, into that name.

PR #499, opened 2020-03-25
feat(selections): adding workbench selections and actions

ramfox:
added array of selections that we will need for the coming container refactor

added `writeDataset` function that combines `fsiWrite`, `setMutationsDataset`, and `setMutationsStatus`, adapted from the `handleSetDataset` function in `Workbench`

This pr just adds the functions, it doesn't remove any current functionality from the `Workbench` component

### DIFF
--- a/app/actions/workbench.ts
+++ b/app/actions/workbench.ts
@@ -1,8 +1,12 @@
-import { LaunchedFetchesAction } from '../store/api'
+import { LaunchedFetchesAction, ApiActionThunk } from '../store/api'
+import fnv from 'fnv-plus'
+import cloneDeep from 'clone-deep'
 
 import {
   setCommitTitle,
-  setCommitMessage
+  setCommitMessage,
+  setMutationsStatus,
+  setMutationsDataset
 } from './mutations'
 
 import {
@@ -14,9 +18,13 @@ import {
   fetchHistory,
   fetchStats,
   fetchWorkingDataset,
-  fetchWorkingStatus
+  fetchWorkingStatus,
+  fsiWrite
 } from './api'
 import { setCommit } from './selections'
+import { Dataset } from '../models/dataset'
+import { Status } from '../models/store'
+import { selectWorkingDataset, selectStatusFromMutations } from '../selections/workbench'
 
 // fetchworkBench makes the necessary API requests to populate the workbench
 // based on what we know about the working dataset from the state tree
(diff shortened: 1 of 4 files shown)


app/selections/workbench.ts
```diff
@@ -0,0 +1,53 @@
+import Dataset from "../models/dataset"
```
b5: very into the functions in this file, I'm wondering about it's location. how you feel about [this redux recommendation](https://redux.js.org/faq/code-structure#code-structure):

> It's generally suggested that selectors are defined alongside reducers and exported, and then reused elsewhere (such as in mapStateToProps functions, in async action creators, or sagas) to colocate all the code that knows about the actual shape of the state tree in the reducer files.

We have so much trouble reshaping the state tree, and I'm wondering if this'll help. That said, the `workingDataset` (workbench) reducer file is already rather long.

app/actions/workbench.ts
```diff
@@ -59,3 +67,52 @@ export function fetchWorkbench (): LaunchedFetchesAction {
...
   }
 }
+
+export function writeDataset (peername: string, name: string, writeDataset: Dataset): ApiActionThunk {
+  return async (dispatch, getState) => {
+    const state = getState()
+    const head = selectWorkingDataset(state)
+    const prevStatus = selectStatusFromMutations(state)
+    const newStatus = determineMutationsStatus(head, writeDataset, prevStatus)
+    if (state.workingDataset.fsiPath !== '') await fsiWrite(peername, name, writeDataset)(dispatch, getState)
```
b5: We should always use curly braces on if statements. it's tough to see that this is potentially asynchronous on first glance

app/actions/workbench.ts
```diff
@@ -59,3 +67,52 @@ export function fetchWorkbench (): LaunchedFetchesAction {
...
+    dispatch(setMutationsStatus(newStatus))
+    return dispatch(setMutationsDataset(writeDataset))
+  }
+}
+
+function determineMutationsStatus (head: Dataset, mutation: Dataset, prevStatus: Status): Status {
```
b5: I'd love a comment on what this function does.

app/selections/workbench.ts
b5: having all the state tree stuff in one file would also make for easier testing. Test coverage would be particularly useful

ramfox: I'm into the idea that everything that knows about the shape of the state tree should be near each other, but I was having trouble with this since some of the selections so far also pull from the `mutations` state tree, some from the `workingDataset` tree, and some from the `commitDetails` tree, and I'm not sure where the boundaries should be.

I'm thinking

ramfox: Basically settled on using 'this-function-will-be-used-in-workbench' as a delimiter for this, but not sure that will track as we move forward. aka not sure there should be a `selections/collections.ts` and instead have those selections live in the `myDatasets` reducer

PR #503, opened 2020-03-27
refactor(Workbench): turn dataset components into containers!

### LINKED ISSUE #498: refactor(workbench): refactor children components into containers
realized I was tackling this without an issue to help track:

Welcome to the refactoring workbench issue. We want to utilize selectors and containers to better structure our workbench component!

### renaming & dataset component conventions
Some components have a 'history' and 'status' version. Usually, when they are materially different aka `Commit` & `CommitHistory`, `MetadataEditor` & `Metadata`, `Readme` & `ReadmeHistory`.

Structure switches internally on whether or not to show the editor or the display. Instead, let's enforce, for now, that there are two different versions of each component, a `COMPONENTEditor` and a `COMPONENT`. So let's refactor Structure into `StructureEditor` and `Structure`

DatasetContainer will get a `location` passed from workbench. This will let it know if it should be showing the `Editor` or display versions of each dataset component.

### todos:

* [x] create selectors for datasets (with and without mutations), status (with and without mutations)
* [x] create `writeWorkingDataset` action that pulls out write calls into one action that can be used for each component container
* [x] rename `Commit` to `CommitEditor` & rename `CommitHistory` to `Commit`
* [x] rename `Readme` to `ReadmeEditor` & rename `ReadmeHistory` to `Readme`
* [x] refactor `ReadmeEditor` into container
* [x] refactor `Readme` into container
* [x] rename `Structure` folder to `Schema`
* [x] rename `FormatConfigFSI` to `FormatConfigEditor` and move into `workbench`
* [x] rename `FormatConfigHistory` to `FormatConfig` and move into `workbench`
* [x] move `Structure.tsx` into `workbench` 
* [x] add `StructureEditor` (container) and refactor out editing from `Structure.tsx` & turn into a container
* [x] refactor `MetadataEditor` into container
* [x] refactor `Metadata` into container
* [x] refactor `Body` into container
* [x] refactor `DatasetComponent` into container
* [x] refactor `CommitDetails` into container

ramfox:
### refactors

closes #498 

adds to selections

### fixes

* fix(workbench): when fsi true, don't prompt the user about unsaved changes

No need to prompt the user about unsaved changes. But we also need fetch the dataset, so that if the user navigates way and comes back (which won't trigger a re-render, since nothing is different in the workbench), we have all the must recent details.

* fix selectStatusFromMutations to correctly pull or create a status

* fix body pagination that didn't have the correct `fetchBody` or `fetchCommitBody` passed into the `Body` component (fix is a side effect of containerizing `Body`)


app/components/workbench/DatasetComponent.tsx
```diff
@@ -147,15 +124,8 @@ export const DatasetComponent: React.FunctionComponent<DatasetComponentProps> =
...
-              ? <ReadmeHistory />
-              : <Readme
-                data={data.readme}
-                name={name}
-                username={peername}
-                write={handleWrite}
-                isLinked={fsiPath !== ''}
-                loading={isLoading}
-              />}
+              ? <Readme />
```
b5: We should formally require the working dataset reference be passed to all of these components as a prop. The container should use the optional `ownProps` argument to pass it down to the component itself.

Popping open the app & clicking from the collection to the workbench, you get these two actions:

```js
{
  type: 'SELECTIONS_SET_WORKING_DATASET',
  payload: {
    peername: 'b5',
    name: 'covid_19_recovered'
  }
}
```

then:
```js
{
  type: '@@router/LOCATION_CHANGE',
  payload: {
    location: {
      pathname: '/workbench/b5/covid_19_recovered',
      search: '',
      hash: ''
    },
    action: 'PUSH',
    isFirstRendering: false
  }
}
```

The information about what dataset reference we're navigating to (`b5/covid_19_recovered`) is in both actions. We should strive to remove the `SET_WORKING_DATASET` action altogether, favouring the information provided by the current route.

The right way to remove `SET_WORKING_DATASET` is to decouple the reference these sub components are told to look at. Because these components are containers, they can still call actions that do fetching to bloom out from the provided reference, but the reference they act upon needs to come from an outside component.

We don't need to remove `SET_WORKING_DATASET` in this PR, which would mean this dataset component can still just pull the reference from the state tree & pass it to each component. But I do think we should do the work of requiring a reference prop for each component here & all their `Editor` variations (`Readme, Metadata, Structure, etc.`).

This'll make it easier to do "multiple workbench" scenarios, deep linking into the app, and make storybook composition easier. As for the prop itself, I think we should formally model it as a `QriRef` type, using "qriRef" as a prop to disambiguate from react's `ref` prop:

```tsx
<Readme qriRef={ref} />
```

`/app/components/network/network.tsx` and `/app/conatiners/NetworkContainer.tsx` are an example of this pattern in action.

I totally recognize this has been the case before this PR, but we're refactoring! Might as well push for a clean slate to work from.

app/components/workbench/MetadataEditor.tsx
```diff
@@ -15,6 +15,11 @@ import { ApiActionThunk } from '../../store/api'
 import { Meta } from '../../models/dataset'
 
 import SpinnerWithIcon from '../chrome/SpinnerWithIcon'
+import { selectMutationsDataset, selectWorkingDatasetIsLoading } from '../../selections'
+import Store from '../../models/store'
+import { Dispatch, bindActionCreators } from 'redux'
+import { writeDataset } from '../../actions/workbench'
+import { connect } from 'react-redux'
```
b5: import sort

app/components/workbench/Readme.tsx
```diff
@@ -2,21 +2,20 @@ import * as React from 'react'
...
 
 import Store from '../../models/store'
+import { selectHistoryDatasetRef } from '../../selections'
 
-export interface ReadmeHistoryProps {
-  peername: string
-  name: string
-  path: string
+export interface ReadmeProps {
+  datasetRef: string
```
b5: would love it if this switched to a `qriRef` type

app/components/workbench/ReadmeEditor.tsx
```diff
@@ -1,26 +1,30 @@
...
-import SpinnerWithIcon from './chrome/SpinnerWithIcon'
-import { datasetConvertStringToScriptBytes } from '../utils/datasetConvertStringToScriptBytes'
-import Dataset from '../models/dataset'
+import SpinnerWithIcon from '../chrome/SpinnerWithIcon'
+import { datasetConvertStringToScriptBytes } from '../../utils/datasetConvertStringToScriptBytes'
+import Dataset from '../../models/dataset'
+import { selectWorkingDatasetRef, selectWorkingDatasetIsLoading, selectMutationsDataset, selectIsLinked } from '../../selections'
+import Store from '../../models/store'
+import { writeDataset } from '../../actions/workbench'
+import { Dispatch, bindActionCreators } from 'redux'
```
b5: import sort

app/components/workbench/ReadmeEditor.tsx
```diff
@@ -1,26 +1,30 @@
...
 
 const DEBOUNCE_TIMER = 1000
 
-export interface ReadmeProps {
-  username: string
-  name: string
+export interface ReadmeEditorProps {
+  datasetRef: string
```
b5: switch type to `qriRef`

app/components/workbench/Transform.tsx
```diff
@@ -1,34 +1,30 @@
 import * as React from 'react'
 import { connect } from 'react-redux'
-import { bindActionCreators, Dispatch } from 'redux'
+import { ApiActionThunk } from '../../store/api'
```
b5: import sort

app/components/workbench/Workbench.tsx
```diff
@@ -178,85 +165,6 @@ class Workbench extends React.Component<WorkbenchProps, Status> {
     this.props.discardMutationsChanges(component)
   }
 
-  determineMutationsStatus (head: Dataset, mutation: Dataset, prevStatus: Status): Status {
```
b5: wow this cleanup is bonkers. Nice work

app/components/workbench/Commit.tsx
```diff
@@ -1,135 +1,32 @@
...
-    if (isValid) {
-      saveWorkingDatasetAndFetch()
-        .then(() => {
-          setCommitTitle('')
-          setCommitMessage('')
-        })
-    }
-  }
-
+export const CommitComponent: React.FunctionComponent<CommitProps> = ({
```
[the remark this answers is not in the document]
ramfox: think we should go for eyes over hands, especially from the perspective of getting folks to contribute. Easier for them to understand what's going on if it says `FunctionComponent` over `FC`

PR #519, opened 2020-04-02
refactor(App, modals): containerize, clean up modals and App

ramfox:
- add `dismissModal` action
- containerize all modals, modals pull only the `modal` object from the state tree. This object is added to the state
 tree by the `setModal` function. The only params a modal is allowed to pull from the state tree is any information t
hat is persisted in local storage, otherwise all params should be sent to the state tree in the `setModal` function.
- removed duplicate `Toast` component, containerized the de-duplicated `Toast` component
- containerized `Navbar` component
- removed `AppContainer`
- removed now unnecessary props from `App`, slimming it down significantly


app/components/workbench/Workbench.tsx
```diff
@@ -133,8 +133,16 @@ export class WorkbenchComponent extends React.Component<WorkbenchProps> {
...
+      ? setModal({
+        type: ModalType.UnpublishDataset,
+        username: this.props.qriRef.username,
+        name: this.props.qriRef.name
+      })
+      : setModal({
+        type: ModalType.PublishDataset,
+        username: this.props.qriRef.username,
+        name: this.props.qriRef.name
+      })
```
b5, line 145: In a future PR we should strive to introduce stronger typing here. We have a few options:
* define `showModal` actions that specify required arguments for each modal type. eg: `showModalUnpublish(username: string, name: string)`
* import various `Modal` data types & use assertions on the generic `setModal` action
* do a combination of both: change `setModal` to require one of the `Modal` datatypes as a second argument

PR #525, opened 2020-04-10
refactor(workbench): add routes to Workbench

### LINKED ISSUE #524: refactor(workbench): use nested routes to control the content on the workbench
Eventually, we will be refactoring the `Workbench` into the `Collection` page. We also want to get off of our reliance on the `selections` state tree. To reach both these goals, let's refactor the workbench into a series of nested routers.

First, the workbench should route between `edit` (`/workbench/edit/:username/:name`) mode and `history` (`/workbench/:peername/:name`) mode. Then, the `DatasetComponent` container should be refactored into a nested router that switches between components.

## Workbench Router

Create components that each use `Layout`! Biggest change here is that the logic for the sidebar will now live along with the main content. This is a win because now instead of having to worry about the state of the world in two different places that care about the same information, we now have one! Also, its easier to reason about what we expect each route should show if the sidebar and main content are in the same location.

each route needs to check if the user `hasDatasets`
if not, redirect to
**/workbench/nodatasets**
- No Datasets
  - sidebar -> disabled WorkingComponentList
  - main -> current NoDatasets content

Potential avenue for NoDatasets => Create a `WorkbenchRoute` component that wraps/extends Route, that checks `hasDatasets` and redirects if the user doesn't have datasets. `hasDatasets` should only be false if `myDatasets.pageInfo.isFetching === false && myDatasets.value.length === 0`.

**exact /workbench**
- No dataset selected
  - sidebar -> disabled WorkingComponentList
  - main -> current NoDatasetSelected content

**/workbench/edit**
- Edit Dataset
  - sidebar -> WorkingComponentList
  - main -> DatasetComponent

**/workbench/:username/:name/at/ipfs/:path**
- History Dataset
  - sidebar -> WorkbenchLogList
  - main -> CommitDetails

**/workbench/:peername/:name**
- if no history, redirects to edit
- if history, redirects to /workbench/:peername/:name/at/ipfs/:path

### pathTo functions
We should have uri builders for each path (like selectors: functions that centralize the logic of creating each path) that we can use when we `push`, so that if/when we adjust routing, there is only one location we have to adjust in response to get our `Link` components and `history.push` functions to function properly.
Right now, each starts with `pathTo`... but open to better naming if suggested.
  - pathToNoDatasets()
  - pathToNoDatasetSelected()
  - pathToEditDataset(username, name)
  - pathToEditDatasetComponent(username, name, component)
  - pathToHistoryDataset(username, name, path?)
  - pathToHistoryComponent(username, name, path, component)
  - pathToWorkbench()
  - pathToCollection()
  - pathToNetwork()
  - pathToNetworkDataset(username, name, path?)
  - pathToCompare()
  - pathToNetworkProfile(username)

### qriRef questions/notes
- NOTE: qriRefs need to be re-built at each level of the router. eg, a qriRef that exists just below the main router but above the workbench router will not have the `path` param, a qriRef that exists below the workbench router but above the components router will not have the `selectors` param. Trying to understand what should dictate passing 

- Q for b5. How do you feel about us renaming `qriRef.selectors` to `qriRef.component` on frontend? trying to figure out/understand how we might use selectors in the future. There is an unbounded length to the depth the selectors can try and access. Do we keep `selectors` and parse `selectors` when we want to get the component? or do we have `component` be its own param, along side `selectors`, which indexes into a component? since you need to specify a component in order to index into it (after the dataset level, the first index in will have to be a component), this seems natural: that a component should have its own param. It would also just make our lives easier on the frontend. Once we start allowing folks to link to specific cells or fields in a dataset component, then we can figure out how to best parse `selectors`

### Component Router (refactor of DatasetComponent)
`${url}/error`
  - covers case when there is a `parseError` in status. This is like the above `hasDatasets === false` situation above... should research this, this feels like a problem that has to come up & probs has a reasonable solution beyond the one I've proposed.
`${url}/meta`
  - default?
  - switch on path includes edit???
  - meta component vs metaEditor component
`${url}/body`
 - body
`${url}/commit`
  - switch on path includes edit?
  - commit component vs commitEditor component
`${url}/transform`
  - switch on path includes edit?
  - transform
`${url}/readme`
  - switch on path includes edit??
  - readme component vs readmeEditor component

The `path` won't contain the filled in params, so we don't have to worry about if a dataset has `/edit` in it somehow

Another option is to have a ComponentRouter and an EditorComponentRouter... but I don't feel good about that. It is nice to have similar logic in the same spot for now.

### Adding back selections
We still want folks to be able to navigate away and to components and have the same dataset and the same location waiting for them, this is the expected behavior of many apps.

Let's add a 'workbenchRoutes' reducer that contains `historyRef`, `editRef`, and `workbenchLocation`

When we have a route change within the Workbench, add the most recent qriRef to either the `historyRef` or `editRef`, as well as populating the `workbenchLocation` with the corresponding location string.

We need to add the 'last' location route, and push to that location at certain places:

1) clicking the 'workbench' link in the nav bar should navigate to `workbenchLocation`
2) clicking 'status' => should navigate back to the `editRef`
3) clicking 'history'=> should navigate back to the last `historyRef`

ramfox:
## Massive PR adding routing to the Workbench, removing selectors

The main objective in this PR is to remove our reliance on 'selectors' and instead control what we display in app using routing.

A secondary objective to is to make future refactors easier. First, by encapulating logic for navigating and selecting from the state tree. Second, by distangling our reliance in actions on the state tree.

Finally, there was ALOT of renaming of components, parameters, and shuffling around of files.

### api action changes
Api action functions no longer rely on parameters from the state tree to fetch from the backend. All `username` and `name` parameters are passed in. There are still a few places where we are relying on state in order to fetch, but this percentage has vastly lowered, and all cases are now wrapped in selector functions, rather then indexing into state directly.

### selections to qriRef & the router
We no longer rely on selections for navigation, relying solely on the router, and our 
There are one or two places where we are cheating a little bit, in both cases the parent component needs to know what component is currently selected. If the component param isn't available at that point in the route tree, we derive the selected component from the location string.
We are heavily dependant on `qriRef` to understand what any component should be displaying. Almost at every point in the `Workbench` component tree we generate a `qriRef` from RouteComponentProps, either because the RouteComponentProps are passed in from the Router directly, or we wrap the component in a `withRouter`.

### layout
The Workbench now has a few `Layout` components that allow for pages to have the same format. `WorkbenchLayout` wraps the `Layout` component, and wraps the child `mainComponent` and `sidebarComponent` in `WorkbenchMainComponent` and `WorkbenchSidebarComponent`, that gather the common styling/functionality that belongs on each page of the Workbench.

### workbenchRoutes reducer
In order to keep track of the last page the user has visited, we now have an EXTREMELY paired down version of `selections` called `workbenchRoutes` (which if we expand to keep track of other pages, can be renamed to `latestRoutes`). This reducer keeps track of the last `editRef` and `historyRef`, to be used when switching back and forth between `status` and `history`, as well as from other views to the workbench, in general.

### renaming components, parameters, and files
We are about 85% of the way to only using the term `username`, rather then `peername`, in the frontend. The api still expects `peername` in it's responses.

CommitDetails has been changed to `HistoryDataset` to parallel `WorkingDataset`/`EditDataset` (we osillate between refering to the editable dataset as `working` and `edit`, which we should settle on)

We now have a `layouts` folder in `workbench` that contain the different layout components
We now have a `components` folder in `workbench` that contains all the dataset componet components, both the display and editor versions

`Log` is now being used where `History` was used before (to reference a list of logs). `Log` now has its own reducer, and was removed from the `WorkingDataset` reducer.

### `pathTo` functions
Any time we `push` to a new route, we use a `pathTo` function. `pathToEditDataset` `pathToHistoryDataset`, `pathToCollection`, `pathToNetworkDataset`, `pathToWorkbench`, `pathToNetwork`, all take the needed params and return a uri string that will send us to the correct location. Now when/if we refactor the router there is one function to adjust.

### router
There are two main routers in workbench, the `WorkbenchRouter` and `ComponentRouter`.

The WorkbenchRouter switches between:
- exact path (`/workbench`) => shows NoDatasets component
- `${path}/edit/:username/:name` => EditDataset component (if the user has no datasets, redirects to the `/workbench` endpoint)
- ${path}/:username/:name/at/ipfs/:path` => HistoryDataset component (if the user has no datasets, redirects to the `/workbench` endpoint)
- `${path}/:username/:name` => redirects to `${path}/edit/:username/:name`

The ComponentRouter switches between:
- exact path (no component selected) => redirects to `${path}/body`
- `${path}/meta`
- `${path}/body`
- `${path}/readme`
- `${path}/structure`
- `${path}/transform`
Each route checks if the path is an `edit` path, or not, and returns the correct component
We moved checking for parsing errors in the status into each Edit component itself.

### general bug fix
- in `DatasetReference`, if a dataset is not in your namespace, you shouldn't be able to rename it
- in `DatasetItem`, use the path in the `myDatasets` list to navigate directly to the latest version of the dataset, when you click on a dataset in the collection

closes  #524


app/components/Navbar.tsx
```diff
@@ -117,7 +118,7 @@ export const NavbarComponent: React.FunctionComponent<NavbarProps> = (props: Nav
     {
       icon: faFileAlt,
       id: 'workbench',
-      link: '/workbench',
+      link: recentWorkbenchLocation === '' ? '/workbench' : recentWorkbenchLocation,
```
b5, line 121: if `selectRecentWorkbenchLocation` defaulted to returning `/workbench` instead of the empty string we can rename the interface prop to just `workbenchLink` and pass it directly to here.

PR #531, opened 2020-04-20
fix(workbench): fix small reducer and fetching bugs 

ramfox:


app/reducers/workingDataset.ts
```diff
@@ -85,7 +85,7 @@ const workingDatasetsReducer: Reducer = (state = initialState, action: AnyAction
         isLoading: false,
         components: {
           readme: {
-            value: dataset && dataset.readme ? atob(dataset.readme.scriptBytes) : undefined
+            value: dataset && dataset.readme && dataset.readme.scriptBytes ? atob(dataset.readme.scriptBytes) : undefined
```
b5, line 88: I'm not a fan of using the ternary operator to return undefined, ever.

Yes, this is a crime I myself have committed in the past. In the history of this repo even.

ramfox: adding to list of refactors

PR #533, opened 2020-04-22
refactor(*): abstract `RouteComponentProps` and `connect`, rename `HistoryDataset`, clean up ternary operators

ramfox:
In a bid to abstract dependency code and reduce imports, here is `connectComponentToProps` and `connectComponentToPropsWithRouter`.

Pass in a component, an optional mapStateToProps functions (or an object with props) and an optional mapDispatchToProps functions (or an object with actions) to get a connected component!


app/utils/connectComponentToProps.ts
```diff
@@ -0,0 +1,30 @@
+import * as React from 'react'
+import { Dispatch, bindActionCreators } from 'redux'
+import { connect } from 'react-redux'
+import { withRouter } from 'react-router-dom'
+
+import Store from '../models/store'
+
+// want to return two objects,
```
b5, line 8: this comment isn't doing much

app/utils/connectComponentToProps.ts
```diff
@@ -0,0 +1,30 @@
...
+export function connectComponentToProps<T> (mapStateToPropsFunc: (state: Store, ownProps: T) => T, mapDispatchToPropsFunc: (ownProps?: T) => any, component: React.FunctionComponent | React.ComponentClass, needHistory?: boolean) {
```
b5, line 9: The use of a type variable here is very nice. Generally, generics are nice 😄 . 

I'm Not a fan of the `needHistory` boolean arg, makes the call site less readable. I think it should be broken out into a separate function, also exported from this file:
```ts
export function connectComponentToPropsWithRouter<T> (
  mapStateToPropsFunc: (state: Store, ownProps: T) => T, 
  mapDispatchToPropsFunc: (ownProps?: T) => any, 
  component: React.FunctionComponent | React.ComponentClass
) {
  // ...
}
```

I'm also a fan of breaking out args this complicated to multiple lines. Sad reality when dealing with function signatures that are this complex.

app/utils/connectComponentToProps.ts
ramfox: HAHAH

app/utils/connectComponentToProps.ts
ramfox: perf ty

app/actions/api.ts
```diff
@@ -18,16 +18,16 @@ import {
 import { getActionType } from '../utils/actionType'
 import { datasetConvertStringToScriptBytes } from '../utils/datasetConvertStringToScriptBytes'
 
-import { CLEAR_DATASET_HEAD } from '../reducers/historyDataset'
+import { CLEAR_DATASET_HEAD } from '../reducers/dataset'
 import Dataset from '../models/dataset'
 
-import { pathToHistory, pathToCollection } from '../paths'
+import { pathToDataset, pathToCollection } from '../paths'
```
b5, line 24: right out the gates, the change to just `dataset` is making reading easier

app/actions/mappingFuncs.ts
```diff
@@ -42,7 +42,10 @@ export function mapHistory (data: any[]): VersionInfo[] {
 // versionInfoToDataset converts a detailed ref to a sparsely-populated
 // dataset object without using any fetching
 export function versionInfoToDataset (vi: VersionInfo): Dataset {
-  const theme = (vi.themeList) ? vi.themeList.split(',', -1) : undefined
+  let theme
+  if (vi.themeList) {
+    theme = vi.themeList.split(',', -1)
+  }
```
b5, line 48: better. ty

PR #561, opened 2020-07-08
fix(Storybook): re-enable Storybook to build

### LINKED ISSUE #557: bug(Storybook): stories which render components that import from electron error out
**Overview of the Issue** 
- When `stories` which import from the `electron` package are loaded, the webpack build fails with an error referencing how `fs` cannot be found by electron

Screen Shot 2020-06-29 at 12 42 40 PM

**Suggest a Fix** 
- We need 2 file versions of each affected component -- one for [election](https://github.com/qri-io/frontend/blob/master/lib/components/ExternalLink.electron.js) and one for [web](https://github.com/qri-io/frontend/blob/master/lib/components/ExternalLink.web.js) (examples linked)
- We need to update webpack.config for both the electron app & storybook to switch on a `TARGET_PLATFORM` env var which points to the correct file above so Storybook isn't trying to access `fs` in-browser

uhLeeshUh:
closes #557 

- create `TARGET_PLATFORM` env var (for which values are either `electron` or `web`) which is resolved by default in webpack config files but could be custom set in an npm script
- create 'electron' and 'web' versions of all files required by Storybook stories which import from the `electron` package, and have a main entrypoint file for each which conditionally loads from them based off `TARGET_PLATFORM`
- add plugin to storybook and base webpack configs to conditionally build affected files based on the `TARGET_PLATFORM` value

**A question for reviewers:**

While this branch currently gets Storybook to build successfully, I've noticed that some of the stories still fail to render with their own errors thrown in the console (see attached for NavBar). I'm happy to work on fixing these too, but wasn't sure if that should be included in the scope of this issue (since I interpreted it as getting Storybook working again)? Would love some guidance here of if you'd all like me tackle this now or make a separate issue. Affected stories are:

     - DataSet --> Details Subtext
     - Navbar --> Standard
     - Modals --> Search

Screen Shot 2020-07-08 at 2 13 15 PM


package.json
```diff
@@ -213,7 +213,7 @@
     "@babel/core": "7.8.4",
     "@babel/plugin-proposal-class-properties": "7.8.3",
     "@babel/plugin-proposal-decorators": "7.8.3",
-    "@babel/preset-env": "7.8.4",
+    "@babel/preset-env": "7.8.7",
```
uhLeeshUh, line 216: I was originally seeing some babel errors, not the `fs` error mentioned in the issue, and after some googling found that updating this `preset-env` dependency would resolve them. I'm not sure what our process is for patch updates, though? And also a little surprised that other folks have just directly been getting the `fs` error? Want to call this out to see if I'm missing something/make sure I'm not doing an unnecessary update

webpack.config.base.js
```diff
@@ -3,10 +3,13 @@
  */
 
 const path = require('path');
+const webpack = require('webpack')
 const {
   dependencies: externals
```
uhLeeshUh, line 8: Just out of my own curiosity, why do we use externals here? I read the webpack docs on them and it suggests that you'd do this if you expect the modules will be available elsewhere to the client at runtime. Where is that the case for us?

app/components/onboard/WelcomeTemplate.web.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react'
+import WelcomeTemplate, { WelcomeTemplateProps } from './WelcomeTemplate.base'
+
+const WelcomeTemplateElectron: React.FC<WelcomeTemplateProps> = (props) => {
+  // TODO (uhLeeshUh): change onClick functionality here to match desired behavior
```
uhLeeshUh, line 5: I'll pick up on this next for this ticket. Is the desired behavior here to close the WelcomeTemplate?

package.json
ramfox: We do an "upgrade dependencies" audit before we do a release, or when we've fallen behind. Since desktop hasn't been touched in a while and node packages upgrade like crazy, we are going to be pretty out of date. We are pretty fast and loose with our patch upgrades (meaning, no one would really question if you just upgraded everything that has had a patch release), but any major upgrades, especially with electron, react, babel, redux, aka the heavy hitters, we try to be more measured and careful about.

If the upgrading happens like this, when you organically find something wrong and see the fix is upgrading a dep, exactly what you are doing is great. Upgrade it and make a note in the commit.

For the second part of your question, can I get a clarification. Are you saying that folks have been getting the `fs` error that we have been seeing, but without the circumstances we are in (trying to import an electron package when we aren't using electron), or there is an additional error that is being fixed?

uhLeeshUh: Oh ok, great! Good to know about dep upgrades

And sorry to further clarify the second part of the question -- I thought I remembered when b5 was showing us the Storybook issue live during my onboarding call that he was getting the `fs` error that I posted in this issue (and wasn't seeing the `preset-env` babel one that I originally saw). I could be wrong there, though. I was just a little confused as to how that could be possible if that is what I saw

webpack.config.base.js
ramfox: Great find! This is actually unnecessary, but was in there based on the boiler plate used b/c of how we structured our app.  

At the time of creating our electron/react app, it was 'best practices' to place your electron code in its own 'app' folder. (Not sure if how we are operating now is considered outdated). If we had any dependencies the were unique to the electron code, it could have gone into its own `package.json`. This call would expose those dependencies. But... we don't actually use the `dependencies` field in the `./app/package.json` file, so again this is actually unnecessary!

package.json
ramfox: Ahhhh i see now! okay, I wouldn't worry too much, could be a node version conflict!

uhLeeshUh: aha! ok good to know, I hadn't thought about node version

webpack.config.base.js
uhLeeshUh: ah ok, thanks, good to know! I had glanced over the `/app` in the `package.json` import so that makes more sense now 👍 Do you want me to remove it if it's unneeded and I'm working on the file? Or do you all typically handle that sort of thing in tech debt tickets?

ramfox:
Okay this is wonderful! Pumped we have storybook working!

Don't hate me forever, but I think the way we structure the imports/base files/targets should be rearranged :)

In the current version as a dev, you have to remember/know that the component you are importing has a TARGET_PLATFORM. And then the `base.ts` file doesn't end up enumerating at a glance all the content you could be trying to get at. I think we should rearrange where we still have a file that contains all the content we need to know about, so it's easier to update, use, and for new folks to grok.

This also means we are containing the number of places we see `TARGET_PLATFORM`, which can throw off new devs. I think keeping it a bit hidden is okay until someone actually has to do some work requiring them to understand what's going on.

Think we should change it to this format, using `/actions/ui.ts` as an example:
- move `ui.electron.ts` and `ui.web.ts` into a `ui` folder as you mentioned previously.
- keep the main file with the un-adorned name aka '/actions/ui.ts' (rather then '/actions/ui.base.ts'). This is the file you import anytime you want to use any UI actions
- inside '/actions/ui.ts',  we export all the platform specific functions by name. In this case:
```
// signout has platform specific implementation
// when using the `signout` function, import it from this file
export { signout } from './ui/ui.TARGET_PLATFORM'
```
- re-change the imports from `/actions/ui.base.ts` to `/actions/ui`
- remove `/actions/ui.base.ts`

What do you think? I'm wavering between hiding the `.electron.ts` and `.web.ts` files in their own folders and not, and what the folder should then be called.

folder name options:
/ui/ui.electron.ts
/ui/ui.web.ts
w/ export looking like 
`export { signout } from './ui/ui.TARGET_PLATFORM'`

and

/platformSpecific/ui.electron.ts
/platformSpecific/ui.web.ts
`export { signout } from '/platformSpecific/ui.TARGET_PLATFORM`

uhLeeshUh:
@ramfox haha I certainly will not hate you forever 🤣  Thanks for that comment! I _really_ like the idea of inverting the file structure so that one file deals with conditional imports and abstracts away a lot of the confusion of what `TARGET_PLATFORM` is and why it's needed. That will make things a lot clearer down the road! Happy to get started on making that change.

And re: the file structure, I do think I'm a fan of hiding the .electron and .web files into their own folders. I actually like the explicit `platformSpecific` name to be really clear about what's going on

app/components/onboard/WelcomeTemplate.web.tsx
ramfox: I think desired behavior might be to... perhaps redirect to the main `qri.io` website?

Other arguments:
- replace "exit" with "back" & have go to the previous page
- have no "exit" text and assume the user can navigate the site on their own

ramfox: actually i've thought about it for a little more, and I believe if we ever pull this component out to use in in a browser, there will be more product design thought that goes into it. I'd say, have "exit" just send the user to the main qri.io page, and don't worry about it beyond that.

uhLeeshUh:
@ramfox ready for a re-review by you! I made the changes you recommended re: have a main file handle the work of importing for the `TARGET_PLATFORM` suffixed files when needed and abstract that logic away from devs. When I pick up on Friday, I'll get to that TODO for the `WelcomeTemplate`. In the meantime, appreciate any other feedback you have / your thoughts on how to handle the components not rendering in Storybook!

app/components/platformSpecific/WelcomeTemplate.web.tsx
```diff
@@ -0,0 +1,3 @@
+export const exitLinkAttributes = {
+  href: 'https://qri.io'
```
uhLeeshUh, line 2: I saw when poking around that we have this [constants file](https://github.com/qri-io/desktop/blob/master/app/constants.js) and that we also make reference to the main qri.io site a few times in the code base. Is it worth it for me to make that into a constant and import it here instead of the hardcoded string?

uhLeeshUh: Also - another note to make here is I made the assumption that if the user is choosing to exit then they don't need to open this site in a different tab (so I didn't adjust the `target: _` attribute) but let me know if my assumption is wrong there

ramfox: I think that's a good assumption.

And yes, good idea re: constants file!

ramfox:
> I I also plan to tackle the failing Circle step on Friday (I'm surprised to see anything I did would affect qri backend or temp_registry_server from downloading? If you have quick thoughts on that as a place for me to start from, that'd be great too)

Ahh, so after taking a look. There are going to be some odd problems here. circleci always builds from HEAD for qri and temp_registry_server, and head is going to break/cause problems here.

Don't worry about circleci for now. Are the tests passing locally?

uhLeeshUh:
> Ahh, so after taking a look. There are going to be some odd problems here. circleci always builds from HEAD for `qri` and at latest release for `temp_registry_server`, which are going to have breaking changes.
> 
> Don't worry about circleci for now. Are the tests passing locally?

@ramfox ahh I see. The unit tests are passing and all e2e tests are passing except the one that was failing for me last week (`clicking 'Network' tab takes you to the feed, navigate to network preview, new commits are foreign`)

uhLeeshUh:
> I agree with your interpretation of the scope of this pr. Can you open an issue describing each issue with the components that aren't rendering?
@ramfox that issue is #564

PR #565, opened 2020-07-13
fix(Storybook): get all Storybook stories to render

### LINKED ISSUE #564: bug(Storybook): fix stories which error on render
**Overview of the issue:**
Currently the following stories error when running Storybook:

1. `LineDiff: 'LineDiff: +1-3'`: there is a TypeError for an `inserts` property that is undefined at runtime. Will likely involve stronger typing and handling the undefined case in code
Screen Shot 2020-07-10 at 1 28 26 PM

2. `Table: 'basic mixed edits', 'column swap', 'column swap & edits'`: all 3 stories have the same issue with `reduce` not being a property of `undefined`. Likely will involve handling case where passed array is `undefined`
Screen Shot 2020-07-10 at 1 32 03 PM

3. `Dataset: 'details subtext'`: no apparent console errors for this one, but it just renders a blank page
Screen Shot 2020-07-10 at 1 40 05 PM

4. `Chrome: Navbar: 'Standard'`: `store` doesn't appear to be in context for the `NavBar` component. Likely need to examine redux store config for this component.
Screen Shot 2020-07-10 at 1 41 37 PM

5. `Modals: 'Search'`: an error exists for using `withRouter` outside of a `<Router>` component
Screen Shot 2020-07-10 at 2 00 45 PM

6. `Network: 'Home'`: an error exists for using `withRouter` outside of a `<Router>` component
Screen Shot 2020-07-10 at 2 00 54 PM

uhLeeshUh:
closes #564 

- Refactors stories which rely on an outdated API
- Utilizes "uncontainerized" versions of components to prevent React Router & Redux errors

### DIFF
--- a/app/components/compare/DiffStat.tsx
+++ b/app/components/compare/DiffStat.tsx
@@ -2,7 +2,9 @@ import React from 'react'
 
 export interface DiffStatProps {
   data: {
+    // inserts may be undefined if the datasets are identical
     inserts?: number
+    // deletes may be undefined if the datasets are identical
     deletes?: number
   }
 }
--- a/app/components/compare/DiffStat.tsx
+++ b/app/components/compare/DiffStat.tsx
@@ -2,8 +2,8 @@ import React from 'react'
 
 export interface DiffStatProps {
   data: {
-    inserts: number
-    deletes: number
+    inserts?: number
+    deletes?: number
   }
 }
 
--- a/stories/6-TableDiff.stories.tsx
+++ b/stories/6-TableDiff.stories.tsx
@@ -19,7 +19,7 @@ const patchStructuredData = {
     "+++": "djs_edits.csv",
     "add": 8, "rem": 4, "mod": 1
   },
-  "patch": [
+  "schema": [
     [[ [1,3], [1,5] ], [ "dj dj booth", 1,"yes" ] ],
     [" ", [ "mafintosh", 3, "no" ]],
     [" ", [ "lindsey losam", 2, "yes" ]],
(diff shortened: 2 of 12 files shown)


stories/6-TableDiff.stories.tsx
```diff
@@ -34,7 +34,14 @@ const patchStructuredData = {
     [" ", [ "com truise", 4, "yes" ]],
     [" ", [ "Ryan Hemsworth",4,"yes"]],
     ["+", [ "Susan Collins", 3, "no"]]
-  ]
+  ], 
+  "diff": [],
```
uhLeeshUh, line 38: Note: for this file, our API now relies on having these `diff` and `stat` fields. I did my best to mock them out but let me know if the data I added doesn't make sense with the `schema`!

stories/Network.stories.tsx
```diff
@@ -1,16 +1,13 @@
 import React from 'react'
-import { BrowserRouter as Router } from 'react-router-dom'
+import { BrowserRouter as Router, Route } from 'react-router-dom'
 
-import feed from './data/splash'
+// import { Dataset as IDataset } from '../app/models/dataset'
```
uhLeeshUh, line 4: Note: I commented these out because they are currently referencing other commented out code in this file and are otherwise unused

stories/6-TableDiff.stories.tsx
ramfox: okay cool, so I need to do a deep dive into this when I finish up this review tomorrow

ramfox:
This is great! I have to do a deeper dive on the Diff expected params, but all the other changes to the storybook components are 👍

app/components/compare/DiffStat.tsx
```diff
@@ -2,8 +2,8 @@ import React from 'react'
 
 export interface DiffStatProps {
   data: {
-    inserts: number
-    deletes: number
+    inserts?: number
```
uhLeeshUh, line 5: When I was playing around with the differ I tried comparing the same file to itself and saw that these fields were undefined, so I went ahead and updated the type here just to cover that case. Do you think this is superfluous?

ramfox: this is good! Would be good to have a notation 
`// inserts may be undefined if the datasets are identical`
`// deletes may be undefined if the datasets are identical`

PR #569, opened 2020-07-15
Migration components

### LINKED ISSUE #567: feat(components): `IncompatibleBackend`, `MigratingBackend`, `MigrationFailed` components
We need three new components!

They should each the `WelcomeTemplate` component.

* [x] `IncompatibleBackend` - props `{ incompatibleVersion: string }`. Should import the expected backend version number (from `version.js`), and says something to the effect of "Incorrect backend version supplied. Expected qri v${expectedVersion}, but have ${incompatibleVersion}. Please file a issue here."
* [x] `MigratingBackend` - "Migrating the Qri backend! This might take a minute..." W/ a loading spinner
* [x] `MigrationFailed` - "The Qri backend migration failed! Try reloading the app, or send a crash report to the Qri team by etc"

uhLeeshUh:
closes #567 

Builds the following new components for use in Desktop's handling of the backend qri migration:
1. `IncompatibleBackend`
2. `MigratingBackend`
3. `MigrationFailed`

Also builds new `ReportCrashAndReload` component to modulate functionality which can be shared by existing `ErrorHandler` component and new `MigrationFailed` one

### DIFF
--- a/app/components/CrashReportSender.tsx
+++ b/app/components/CrashReportSender.tsx
@@ -1,0 +1,90 @@
+import React, { useState } from 'react'
+import Button from './chrome/Button'
+import ExternalLink from './ExternalLink'
+import WelcomeTemplate from './onboard/WelcomeTemplate'
+import { FetchOptions } from '../store/api'
+import versions from '../../version'
+import { CRASH_REPORTER_URL } from '../constants'
+
+interface CrashReportSenderProps extends ErrorProps {
+  title: string
+}
+
+interface ErrorProps {
+  error?: Error
+  errorInfo?: React.ErrorInfo
+}
+
+const CrashReportSender: React.FC<CrashReportSenderProps> = ({ title, error, errorInfo }) => {
+  const [sendingReport, setSendingReport] = useState(false)
+  const [sentReport, setSentReport] = useState(false)
+
+  const handleSendCrashReport = async () => {
+    setSendingReport(true)
+    try {
+      console.log('sending report')
+      await postCrashReport(error, errorInfo)
+      setSendingReport(true)
+      setSentReport(true)
+    } catch (e) {
+      console.log(e)
+      setSendingReport(false)
+    }
+  }
+
+  const handleReload = () => {
+    window.location.hash = '/'
+    window.location.reload()
+  }
+
+  return (
+    <WelcomeTemplate title={title}>
+      <div className="reporting actions">
+        <p>Apologies, Qri desktop encountered an error. Send us a crash report!</p>
+        {sentReport
+          ? <p>Thanks!</p>
+          : <Button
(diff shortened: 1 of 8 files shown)


app/components/CrashReportSender.tsx
```diff
@@ -0,0 +1,90 @@
+import React, { useState } from 'react'
```
uhLeeshUh, line 1: Just a general question here -- what do you think of the name of this component? It technically handles sending the crash report AND reloading the app so I don't love this but couldn't think of anything better. I decided to refactor the functionality for both crash reporting and reloading since we use them in `ErrorHandler` and want them for `MigrationFailed` as well

ramfox: `ReportCrashAndReload`? I agree that we should include that it reloads in there.

uhLeeshUh:
If it's useful to see, here are screenshots of what `ErrorHandler` and `MigrationFailed` look like with the refactor
Screen Shot 2020-07-15 at 2 19 11 PM
Screen Shot 2020-07-15 at 2 19 33 PM

app/components/CrashReportSender.tsx
uhLeeshUh: yes I like that much better! Will change now

app/components/IncompatibleBackend.tsx
```diff
@@ -0,0 +1,18 @@
...
+import ExternalLink from './ExternalLink'
+import { backendVersion } from '../../version.js'
+
+interface IncompatibleBackendProps {
+  incompatibleVersion: string
+}
+
+const IncompatibleBackend: React.FC<IncompatibleBackendProps> = ({ incompatibleVersion }) => (
+  <WelcomeTemplate title='Incompatible Backend Version'>
+    <p>{`Incorrect backend version supplied. Expected qri v${backendVersion}, but have ${incompatibleVersion}. Please file an issue `}
```
ramfox: Can we change the language here to read something like

"The lowest supported backend version is ${lowestCompatibleBackend}, but have ..."

`lowestCompatibleBackend` is a var i have in my branch, for can you add a fake global var `lowestCompatibleBackend` into this file w/ a TODO comment above saying this needs to be replaced when `lowestCompatibleBackend` var exists in `backend.js`?
And can the "Please file an issue here" be on a new line? Unless that looks bad in your opinion once you do it.

uhLeeshUh: New line is a good call, and yup refactored it! I did change the "but have" piece slightly since it sounded awkward to me. Let me know if you're ok with how it looks now

uhLeeshUh: Screen Shot 2020-07-15 at 3 37 25 PM

PR #571, opened 2020-07-16
feat(migration): adjust `BackendProcess`, `main.dev.js` & `App.ts` so we can coordinate migrations!

### LINKED ISSUE #566: feat(migrate): have desktop coordinate backend migrations needed to get desktop app to work
Let's use the qri config, and electron's `ipcMain` and `ipcRender` events to coordinate backend migrations.

We can also use `qri version` to ensure the user is running the correct backend (if "-dev" is present, or `qri connect` is already running, we should ignore this)

### why fail based on `qri version` and `config`?
Qri Desktop, in production, has an implicit contract with the Qri Backend that the WebSocket and HTTP api's work in an expected way. We should ensure that, unless the user is using a `dev` version of qri, these expectations are preserved. 

### how to proceed

in `backend.js` file let's add `string lowestCompatibleBackend` and `number lowestCompatibleConfigRevision`.

in `maybeStartup` func, ping the `/health` endpoint. If we get a response, just return! (This ensures that is qri is  already running, we don't interfere)

in `launchProcess` func:

If the qri version on the PATH is lower than the `lowestCompatibleBackend`, we should send an "incompatible-backend" event (with the incompatible version string) to the main process. (Unless the given qri version has a `-dev` version ending, then just proceed with attempting to connect as normal.)

If the given config revision is lower than the expected, send a "migrating backend" event to the main process, and start qri using `qri connect --migration`. If that fails, send a "migration failed" event to the main process.

The main process should handle these "incompatible-backend", "migrating-backend", "migration-failed" events, and use the `window.webContents.send` func to pass these events to the render processes.

The render process should handle these events.

### UI reducer and the render process event handlers

new type in `/reducer/ui.ts`:
```
export type bootupComponent:  'loading' | 'migrating' | 'migrationFailure' | string
```

UI should have a new field `bootupComponent: bootupComponentType`, which by default is 'loading'

### new components

Create three new components:
- MigratingBackend -> informs the user that we are running migrations on qri and this may take a minute
- MigrationFailed -> informs the user that the migration failed and gives instructions on how to send a crash report
- IncompatibleBacken -> displays expected version and version that caused the error

(detailed in separate issue: https://github.com/qri-io/desktop/issues/567)

### App component

The ipcRender process should have a handler that listens for "incompatible-backend", "migrating-backend", and "migration-failed" events. In each case, send an action to update the UI reducer's bootupComponent field. (in the case of "incompatible-backend", set the field to the incompatible version number). These events should be registered when we use the App component.

In `App.tsx`, add a new prop `bootupComponent`, and pass it the UI.bootupComponent. If we are `loading`, switch on `bootupComponentType` (if not 'loading', 'migrating' or 'migrationFailure', default should be `bootupComponent` is the incompatible version number to pass to a `IncompatibleBackend` component).

ramfox:
closes #566 

My journey through this process has been decently well chronicled in the above issue!

My last comment would be, very open to re-structuring `BackendProcess` to take a `window` so that the main function is less unwieldy and is less responsible for coordinating the behavior. However, it might be nice to keep all the emitted events in the `main.development.js` file, so that behavior isn't hidden down in another class. Anyone have strong opinions?

Edit:
Backend is getting a refactor! Now commands won't "auto-trigger" a migration, unless you have the `--migrations` flag. And a command will fail w/ an exit error (`2`) if the config needs a migration. We can use these to remove the desktop's reliance on knowing about the QRI_PATH and keeping track of config revisions!

### DIFF
--- a/app/backend.js
+++ b/app/backend.js
@@ -1,10 +1,14 @@
 const log = require('electron-log')
 const childProcess = require('child_process')
 const fs = require('fs')
+const yaml = require('js-yaml')
 const path = require('path')
 const os = require('os')
+const http = require('http')
 const { dialog } = require('electron')
 
+var lowestCompatibleBackend = [0, 9, 9]
+var lowestCompatibleConfigRevision = 2
 
 // BackendProcess runs the qri backend binary in connected'ed mode, to handle api requests.
 class BackendProcess {
@@ -12,10 +16,26 @@ class BackendProcess {
     this.qriBinPath = null
     this.process = null
     this.debugLogPath = null
+    this.backendVer = null
+    this.configRev = null
 
     // Default to writing to stdout & stderr
     this.out = process.stdout
-    this.err = process.stderr
+    this.in = process.in
+    this.err = process.stderr;
+
+    [
+      'setQriBinPath',
+      'setBackendVer',
+      'setConfigRev',
+      'standardRepoPath',
+      'checkNoActiveBackendProcess',
+      'checkBackendCompatibility',
+      'checkNeedsMigration',
+      'launchProcess'
+    ].forEach((m) => { this[m] = this[m].bind(this) })
+
+
     try {
       // Create a log whose filename contains the current day.
       const nowTime = new Date();
@@ -41,9 +61,13 @@ class BackendProcess {
         detail: err
       })
     }
+    this.setQriBinPath()
+    this.setBackendVer()
+    this.setConfigRev()
   }
 
-  maybeStartup () {
+  // running this function will ensure that a qriBinPath exists
+  setQriBinPath () {
     // In development node, use installed qri binary
     if (process.env.NODE_ENV === 'development') {
       let processResult = childProcess.execSync('which qri');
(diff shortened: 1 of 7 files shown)


app/backend.js
```diff
@@ -74,16 +100,66 @@ class BackendProcess {
...
+      throw new Error("incompatible-backend")
+    }
+  }
+
   launchProcess () {
     try {
-      let processResult = childProcess.execSync(`"${this.qriBinPath}" version`)
-      let qriBinVersion = processResult.toString().trim();
-      this.process = childProcess.spawn(this.qriBinPath, ['connect', '--setup', '--log-all'], { stdio: ['ignore', this.out, this.err] })
+      this.process = childProcess.spawn(this.qriBinPath, ['connect', '--migrate', '--log-all'], { stdio: ['ignore', this.out, this.err] })
```
b5, line 157: plz add `--no-prompt` to the arg list, keep `--setup`!

ramfox: oops! not sure where my mind when for this one, thank you

app/backend.js
```diff
@@ -92,11 +168,26 @@ class BackendProcess {
...
+      throw err
     } else {
       log.warn(`event ${kind} from backend`)
     }
   }
 
+  checkNeedsMigration() {
+    log.info("checking for backend migrations")
+    try {
+      childProcess.execSync(`"${this.qriBinPath}" list`)
```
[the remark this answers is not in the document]
ramfox: yes sorry, you did specifically say this.

PR #572, opened 2020-07-16
refactor(*): changes to get desktop to work with qri v0.9.9/v0.9.10 changes

### LINKED ISSUE #555: fix(rename): adjust rename fetch function to use new expected params (change `new` to `next`)
Otherwise, renames are not working

### LINKED ISSUE #556: bug(websockets): fsi changes in body do not show up in desktop
still investigating, will expand

### LINKED ISSUE #560: refactor(websockets): update to new websocket message api

ramfox:
Now that we can successfully migrate, let's make sure that all the other functionality that needs updating is fixed!

closes #560
closes #555 
closes #556 

make sure e2e tests are running

relies on qri-io/temp_registry_server#4

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2
 jobs:
   build:
     docker:
-      - image: circleci/golang:1.14.1-node-browsers
+      - image: circleci/golang:1.14.6-node-browsers
         environment:
           GO111MODULE: "on"
     working_directory: ~/desktop
--- a/app/actions/api.ts
+++ b/app/actions/api.ts
@@ -486,7 +486,7 @@ export function publishDataset (username: string, name: string): ApiActionThunk 
     try {
       let response: Action
       response = await dispatch(action)
-      await whenOk(fetchWorkingDataset(username, name))(response)
+      await whenOk(fetchLog(username, name))(response)
     } catch (action) {
       throw action
     }
@@ -513,7 +513,7 @@ export function unpublishDataset (username: string, name: string): ApiActionThun
     try {
       let response: Action
       response = await dispatch(action)
-      await whenOk(fetchWorkingDataset(username, name))(response)
+      await whenOk(fetchLog(username, name))(response)
     } catch (action) {
       throw action
     }
--- a/app/actions/api.ts
+++ b/app/actions/api.ts
@@ -711,7 +711,7 @@ export function renameDataset (username: string, name: string, newName: string):
         method: 'POST',
         body: {
           current: `${username}/${name}`,
-          new: `${username}/${newName}`
+          next: `${username}/${newName}`
         }
       }
     }
--- a/app/actions/api.ts
+++ b/app/actions/api.ts
@@ -508,7 +508,8 @@ export function unpublishDataset (username: string, name: string): ApiActionThun
           name
         },
         query: {
-          remote: 'registry'
+          remote: 'registry',
+          all: true
         }
       }
     }
--- a/app/actions/api.ts
+++ b/app/actions/api.ts
@@ -432,12 +432,12 @@ export function saveWorkingDatasetAndFetch (username: string, name: string): Api
   }
 }
 
-export function addDataset (username: string, name: string): ApiActionThunk {
+export function pullDataset (username: string, name: string): ApiActionThunk {
   return async (dispatch) => {
     const action = {
-      type: 'add',
+      type: 'pull',
       [CALL_API]: {
-        endpoint: 'add',
+        endpoint: 'pull',
         method: 'POST',
         segments: {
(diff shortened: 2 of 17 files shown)


ramfox:
Some error is thrown when running the e2e tests, that I believe may have been intro-d now that we've changed the backend launch process.

The e2e tests do their own version of launching the backend, so I will start there to investigate.

ramfox:
Okay looks like that error was fixed after pulling/rebasing from master. 

Next task... looks like desktop isn't registering the websocket events as expected.

ramfox:
Okay, so the bug turned out to be:
- the `TestBackendProcess` `qri connect` would launch
- the App would launch as usually
- the App `BackendProcess` would attempt to see if there were any backend processes already running, however, the `TestBackendProcess` would not be fully "connected" so the `BackendProcess` proceeded as if it needed to launch the backend, but it would fail.
- An odd failure would be thrown to the main process  (however not sure why it wasn't caught in the `.catch`?)

Fixed this by waiting until the TestBackendProcess goes fully online (we ping for a health check until we get a good response. If no health response after 15 attempts, we fail), before we continue to launching the app.

ramfox:
OKAY so got that all straightened out!

The only thing we are missing now is https://github.com/qri-io/temp_registry_server/issues/3

the temp registry server is out of date, we need to update it to get it playing nicely with backend 0.9.10

.circleci/config.yml
```diff
@@ -2,7 +2,11 @@ version: 2
 jobs:
   build:
     docker:
+<<<<<<< HEAD
       - image: circleci/golang:1.14.1-node-browsers
+=======
+      - image: circleci/golang:1.14.6-node-browsers
+>>>>>>> 394d43b... refactor(backend): use expected backendVersion constant from version.js
```
b5, line 9: eep

app/actions/api.ts
```diff
@@ -501,19 +501,22 @@ export function unpublishDataset (username: string, name: string): ApiActionThun
...
       [CALL_API]: {
-        endpoint: 'publish',
+        endpoint: 'remove',
         method: 'DELETE',
         segments: {
           username,
           name
+        },
+        query: {
+          remote: 'registry'
```
b5, line 511: add `all: 'true'` here to closer match the CLI invocation. we may add new functionality in the future that makes behaviour different without the `all` flag. Setting it in advance will keep our intent consistent

app/backend.js
```diff
@@ -1,13 +1,12 @@
...
 const childProcess = require('child_process')
 const fs = require('fs')
-const yaml = require('js-yaml')
 const path = require('path')
 const os = require('os')
 const http = require('http')
 const { dialog } = require('electron')
 
-const lowestCompatibleBackend = [0, 9, 9]
+const { backendVersion: expectedBackendVer } = require('../version')
```
b5, line 9: 👏

app/store/wsMiddleware.ts
```diff
@@ -14,6 +14,17 @@ export const wsConnected = () => ({ type: 'WS_CONNECTED' })
...
+var ETCreatedNewFile = "watchfs:CreatedNewFile"
+// ETModifiedFile is the event for modifying a file
+var ETModifiedFile = "watchfs:ModifiedFile"
+// ETDeletedFile is the event for deleting a file
+var ETDeletedFile = "watchfs:DeletedFile"
+// ETRenamedFolder is the event for renaming a folder
+// var ETRenamedFolder = "watchfs:RenamedFolder"
+// ETRemovedFolder is the event for removing a folder
+// var ETRemovedFolder = "watchfs:RemovedFolder"
+
```
b5, line 27: these should be `const`

app/store/wsMiddleware.ts
```diff
@@ -53,7 +64,7 @@ const socketMiddleware = () => {
         }
         break
       default:
-        console.log('default')
+        console.log(event.type)
```
b5, line 67: can we add some clarity here:
```suggestion
        console.log(`received websocket event: ${event.type}`)
```

app/store/wsMiddleware.ts
```diff
@@ -26,22 +37,22 @@ const socketMiddleware = () => {
...
   }
 
-  const onMessage = (store: Store<IStore>) => (event: MessageEvent) => {
-    const payload = JSON.parse(event.data)
-    switch (payload.Type) {
-      case 'modify':
-      case 'create':
-      case 'remove':
+  const onMessage = (store: Store<IStore>) => (e: MessageEvent) => {
+    const event = JSON.parse(e.data)
```
b5, line 41: we should wrap this in a `try/catch` and log if we catch an error with an early return. Otherwise a bad websocket message will crash desktop

Issue #576, opened 2020-07-20
feat(Collection View): build skeleton of new components and render in Storybook


uhLeeshUh:
**Components to be built out:**
- [x] New fixed-height header running the full width of the content pane with top-level action buttons for "Pull Dataset" and "New Dataset"
- [x] Status component with status glyphs

edit (7/22) deprioritizing:
- Dataset actions menu component
- Collection View table which manages multiselect functionality, column sorting and infinite scroll
- Filter pane

uhLeeshUh:
Note: this work has been de-scoped in favor of prioritizing Qri Cloud work, so only the `Header` and `DatasetStatus` components have been built

Issue #554, opened 2020-06-29
Collection View Refactor


chriswhong:
Improvements for the Collection view will allow the user to quickly understand the status of datasets in their Qri collection, find datasets to work with, and perform dataset-level actions (publish/unpublish, remove, etc).  The collection view is "home base" and is where users will take control of their datasets: pushing, pulling, removing, finding, and just understanding what they have.

This work assumes refactoring of the top-level layout of Qri Desktop, with a fixed-height header running the full width of the content pane.  (The current layout uses split sidebar/content on all views without a unified header)

Major Functionality
- List of Datasets - The list shows all datasets in the Qri collection, and is sortable by clicking the 

- Filtering -  The user can filter by string match or date range of last update (we can add more filters later, these are two simple ones to start with) Filter pane can get more complex as users demand more ways to navigate large collections.  One idea is to have a collapsible advanced filter section.

- Dataset status glyphs - Status glyphs provide a quick visual indicator of published/unpublished, file system integration, and sync.   

- Dataset actions - An ellipses icon at the end of each row shows a context menu with dataset-level actions (remove, duplicated, publish, etc)

- Multiselect - User can select multiple datasets using checkboxes and perform bulk actions.

TODO:
- Indicate how many datasets are currently selected, visually highlight selected datasets in addition to the checkbox 
- Warning UI for remove/publish - These currently exist in the app, but would need to be modified to handle multi-select actions
- Think about keyboard-only navigation and actions
- Syncing UI - clearly show that Qri is pulling new data
- Visual indication that a dataset has "new" data that the user hasn't seen yet

![Qri_Desktop_0_4_0_-_v4_sketch](https://user-images.githubusercontent.com/1833820/86026203-1fd2e200-b9fd-11ea-8396-65728525534b.png)

uhLeeshUh:
@chriswhong thanks for putting this together! I have a couple of questions for you about the design and then also would appreciate any feedback you have on how I'm thinking of breaking this down into parts (I also only focused on the Major Functionality bullets for now if that's ok). Once I get your OK, I can split this issue into a handful of smaller ones 

**Questions I have:**

- When you mention this work assumes the refactoring of the top-level layout of Qri Desktop with a new fixed-height header, are you imagining that's a ticket I should first tackle as a blocker to this?
- To piggyback off the first question, does the new header include the search bar in all views except this one (since I imagine filtering on "All datasets" is effectively the same)?
[...]

PR #578, opened 2020-07-22
feat(Header & DatasetStatus): add new Header and DatasetStatus components

uhLeeshUh:
This PR checks off the first 2 components I enumerated in #576, which builds off designs from #554 

- A new `Header` component to eventually be used across the app but built just for Collection View for now
- A new `DatasetStatus` component to be utilized in the new Collection table view

### DIFF
--- a/app/components/ExternalLink.tsx
+++ b/app/components/ExternalLink.tsx
@@ -2,12 +2,13 @@ import * as React from 'react'
 import { onClick } from './platformSpecific/ExternalLink.TARGET_PLATFORM'
 
 export interface ExternalLinkProps {
-  id: string
+  id?: string
   href: string
   children: React.ReactNode
+  className?: string
 }
 
-const ExternalLink: React.FunctionComponent<ExternalLinkProps> = ({ id, href, children }) =>
-  <a id={id} href={href} onClick={(e) => onClick(e, href)} target='_blank' rel="noopener noreferrer">{children}</a>
+const ExternalLink: React.FunctionComponent<ExternalLinkProps> = ({ id, href, children, className }) =>
+  <a id={id} href={href} onClick={(e) => onClick(e, href)} target='_blank' rel="noopener noreferrer" className={className}>{children}</a>
 
 export default ExternalLink
--- a/app/components/ExternalLink.tsx
+++ b/app/components/ExternalLink.tsx
@@ -6,9 +6,10 @@ export interface ExternalLinkProps {
   href: string
   children: React.ReactNode
   className?: string
+  tooltip?: string
 }
 
-const ExternalLink: React.FunctionComponent<ExternalLinkProps> = ({ id, href, children, className }) =>
-  <a id={id} href={href} onClick={(e) => onClick(e, href)} target='_blank' rel="noopener noreferrer" className={className}>{children}</a>
+const ExternalLink: React.FunctionComponent<ExternalLinkProps> = ({ id, href, children, className, tooltip }) =>
+  <a id={id} href={href} onClick={(e) => onClick(e, href)} target='_blank' rel="noopener noreferrer" className={className} data-tip={tooltip}>{children}</a>
 
 export default ExternalLink
--- a/app/components/Header.tsx
+++ b/app/components/Header.tsx
@@ -15,7 +15,7 @@ interface HeaderProps {
 }
 
 export const Header: React.FC<HeaderProps> = ({ title, setModal }) => (
-  <header className="header">
+  <header className="collection-header">
     <h3 className="header-title">{title}</h3>
     <div className="header-column">
       <HeaderColumnButton
--- a/app/components/Header.tsx
+++ b/app/components/Header.tsx
@@ -1,0 +1,43 @@
+import React from 'react'
+import { faPlus, faDownload } from '@fortawesome/free-solid-svg-icons'
+
+import { Modal, ModalType } from '../models/modals'
+
+import { connectComponentToProps } from '../utils/connectComponentToProps'
+
+import { setModal } from '../actions/ui'
+
+import HeaderColumnButton from './chrome/HeaderColumnButton'
+
+interface HeaderProps {
+  title: string
+
+  // setting action
+  setModal: (modal: Modal) => void
+}
+
+export const Header: React.FC<HeaderProps> = ({ title, setModal }) => (
+  <header className="header">
+    <h3 className="header-title">{title}</h3>
+    <div className="header-column">
+      <HeaderColumnButton
+        id="create-dataset"
+        icon={faPlus}
+        label="Create new Dataset"
+        onClick={() => setModal({ type: ModalType.CreateDataset })}
+      />
+      <HeaderColumnButton
+        id="add-dataset"
+        icon={faDownload}
+        label="Add existing Dataset"
+        onClick={() => setModal({ type: ModalType.AddDataset })}
+      />
+    </div>
+  </header>
+)
+
+export default connectComponentToProps(
+  Header,
+  {},
+  { setModal }
+)
--- a/app/components/Header.tsx
+++ b/app/components/Header.tsx
@@ -2,9 +2,7 @@ import React from 'react'
 import { faPlus, faDownload } from '@fortawesome/free-solid-svg-icons'
 
 import { Modal, ModalType } from '../models/modals'
-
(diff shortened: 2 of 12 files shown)


app/components/ExternalLink.tsx
```diff
@@ -2,12 +2,13 @@ import * as React from 'react'
 import { onClick } from './platformSpecific/ExternalLink.TARGET_PLATFORM'
 
 export interface ExternalLinkProps {
-  id: string
+  id?: string
```
uhLeeshUh, line 5: Note: I only made this optional because I noticed we aren't always providing it as a prop and saw some type errors across the codebase. Happy to add back if folks feel strongly about that, though

app/components/collection-v2/DatasetStatus.tsx
```diff
@@ -0,0 +1,34 @@
...
+  linkedToFilesystem: boolean
+  updatesAvailable: boolean
+
+  // fetching action
+  updateDataset(username: string, name: string): ApiActionThunk
+}
+
+const DatasetStatus: React.FC<DatasetStatusProps> = ({ published, linkedToFilesystem, qriRef: { username, name }, fsiPath, updatesAvailable, updateDataset }) => (
+  <div className="flex-space-between">
+    <ExternalLink href={`${QRI_CLOUD_URL}/${username}/${name}`} className="dataset-link" data-tip="Published">
```
uhLeeshUh, line 22: Question here about our use of tool-tips -- I saw comments about how we utilize `ReactTooltip` via the `data-tip` attribute so I went ahead and added tool tips to these glyphs, but I haven't been able to see them in Storybook. I looked at another place in the app where we follow this pattern (`Modal`) and tried adding a tooltip to the `SearchModal` story but also couldn't see it. Is this a known Storybook problem or am I overlooking something?

app/components/collection-v2/DatasetStatus.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+export interface DatasetStatusProps {
+  qriRef: QriRef
+  fsiPath: string
+  published: boolean
+  linkedToFilesystem: boolean
+  updatesAvailable: boolean
+
+  // fetching action
+  updateDataset(username: string, name: string): ApiActionThunk
```
uhLeeshUh, line 17: Do we currently support this functionality in Desktop? I could have very well missed it, but I didn't see it when I played around with the app locally so I just mocked out what I assume the action passed to this component by a parent dataset row would look like. Happy to update if I've missed something

app/scss/_dataset-status.scss
```diff
@@ -0,0 +1,8 @@
+.dataset-link {
```
uhLeeshUh, line 1: I ended up styling the `link` and `button` differently based off of the default styles that we are provided by the user agent stylesheet. I am just matching the default `button` styling to the links here so all of the glyphs appear the same

app/components/ExternalLink.tsx
b5: Generally, external links should have IDs, but there are situations where they make less sense (eg: a set of list items with external links created by within a loop). Think it makes sense to have this optional

app/components/collection-v2/DatasetStatus.tsx
b5: `ReactTooltip` has a funny API. You might need to add these additional pieces of wiring, both are in the `App.tsx` component:

https://github.com/qri-io/desktop/blob/5f720ad19bfb11121bf4870010bac78787d0bd0f/app/components/App.tsx#L178-L186

https://github.com/qri-io/desktop/blob/5f720ad19bfb11121bf4870010bac78787d0bd0f/app/components/App.tsx#L268-L279

app/components/collection-v2/DatasetStatus.tsx
b5: IIRC this is planned but unfinished work. Think it's great to have it added!

uhLeeshUh: Cool! @b5 do you envision adding that action to be a part of this PR or should I include it in the planned work to "wire-up" these components to our API?

app/components/platformSpecific/DatasetStatus.web.tsx
```diff
@@ -0,0 +1 @@
+export const onClickOpenInFinder = () => undefined
```
b5, line 1: I'd love a comment that clarifies this is necessary for webpack, but the "open in finder" icon won't be used on the web, and so this should never be called.

app/components/collection-v2/DatasetStatus.tsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react'
+import Icon from '../chrome/Icon'
+import ExternalLink from '../ExternalLink'
+import { QriRef } from '../../models/qriRef'
+import { QRI_CLOUD_URL } from '../../constants'
+import { onClickOpenInFinder } from '../platformSpecific/DatasetStatus.TARGET_PLATFORM'
+import { ApiActionThunk } from '../../store/api'
```
b5, line 7: imports should be grouped with a blank line to separate in this order:

```
[module imports]

[non-component local imports]

[component imports]
```

This should be re-worked as:

```suggestion
import React from 'react'

import { QriRef } from '../../models/qriRef'
import { QRI_CLOUD_URL } from '../../constants'
import { onClickOpenInFinder } from '../platformSpecific/DatasetStatus.TARGET_PLATFORM'
import { ApiActionThunk } from '../../store/api'

import Icon from '../chrome/Icon'
import ExternalLink from '../ExternalLink'
```

app/components/Header.tsx
```diff
@@ -0,0 +1,43 @@
+import React from 'react'
+import { faPlus, faDownload } from '@fortawesome/free-solid-svg-icons'
+
+import { Modal, ModalType } from '../models/modals'
+
+import { connectComponentToProps } from '../utils/connectComponentToProps'
+
+import { setModal } from '../actions/ui'
+
+import HeaderColumnButton from './chrome/HeaderColumnButton'
```
b5, line 10: Ditto on the import reorg:

```suggestion
import React from 'react'
import { faPlus, faDownload } from '@fortawesome/free-solid-svg-icons'

import { Modal, ModalType } from '../models/modals'
import { connectComponentToProps } from '../utils/connectComponentToProps'
import { setModal } from '../actions/ui'

import HeaderColumnButton from './chrome/HeaderColumnButton'
```

Having this pattern enforced makes it easier to browse the requirements of a given file

app/scss/_header.scss
```diff
@@ -0,0 +1,41 @@
+.header {
```
b5, line 1: if this is truly styling for _all_ header components then I'm all for it. 

But, if it's at all specific to the collection view, we should aim to scope down the css selector to indicate such a thing. Could do any one of `.collection.header { /*... */` or `#collection .header {`, or `.collection-header`. I'll leave it to desktop folks and existing code to decide which is the right convention to use.

app/components/collection-v2/DatasetStatus.tsx
uhLeeshUh: Worked like a charm, thanks! I also realized I needed to be adding the `data-tip` attribute to the base anchor tag in `ExternalLink`

app/components/collection-v2/DatasetStatus.tsx
uhLeeshUh: Ahh awesome, thanks for clarifying that. Will update

app/scss/_header.scss
uhLeeshUh: good catch! I had meant to call out that the `Header` component is intended to be used app-wide and I'd actually like to refactor it at that point to be rendered once in `App` for all routes and pull its title from the route instead of needing a `title` prop as I have here. I figured that it'd be best to not over-optimize now though since we'd only be using this for the Collection view at present. I like your idea of scoping it down for now and revising later if/when we choose to expand the use of `Header`. I poked around and it looks like our most recent css is in the `0.4.0` folder which tends to follow the `.collection-header` pattern, so I went with that 👍

app/components/platformSpecific/DatasetStatus.web.tsx
uhLeeshUh: Done! I also went back and added for `ExternalLink` so that noop is also not confusing

PR #582, opened 2020-08-06
feat(Bit StatsChart): add local styling to StatsChart Bit component

### LINKED ISSUE #581: Bit Spike: investigate how to best implement shared component styling
The team is using [Bit](https://docs.bit.dev/docs/quick-start) which is a command line tool for versioning and sharing components between consuming projects. In our case, we use it to share React components between Desktop and Cloud. Under the hood, Bit wraps each "component" that you define as its own versioned NPM package which can be imported elsewhere. 

What is the best way to handle styling as we introduce these shared components?

The original plan was to not change the styling pattern of Desktop components which would become shared via Bit (aka keep a global styling pattern) but we ran into a handful of issues using that approach, seen in [this POC PR](https://github.com/qri-io/cloud/pull/368) in the Qri Cloud project.

As a followup to feedback from that POC PR, we need to investigate the best way to implement localized component styling. The goal here is (to quote feedback from that PR):

> Ensure the layout and font sizes all work well for whatever size container the component will be rendered in. Everything else should be controllable using the [consuming] app's CSS.

uhLeeshUh:
closes #581 

Here's my first pass at local styling of Bit components taking into account the feedback you all gave in the previous Cloud PR. This Bit component is called `stats-chart-v1` and a live version of what it looks like in the Bit.dev remote can be found [here](https://bit.dev/qri-io/stats/stats-chart-v1?example=5f2c66be93bf3b0019d198bb) (woohoo! Using component-based styling works much better in Bit.dev than our previous global styling attempt).

**A couple things to note:**

1. I went the CSS modules route that @ramfox suggested so we could locally scope styles to `StatsChart`. The pattern I'm suggesting here, if we choose to follow it, is to re-factor Desktop components that we want to share through Bit into an eponymous directory which has some base `{xComponent}.tsx` file which imports a `{xComponent}.module.scss` file. Any other dependant files of the Bit component will be in that directory as well. I updated our webpack config to only create css modules for the `.module.scss` suffix.
2. I experimented with moving any scss variables used by the new Bit component out of the scss file and into a required React prop which the component accepts. It's the responsibility of the consuming app (either Cloud or Desktop) to provide this prop. The example I'm using here is `fontFamily` which maps to the `$font-family-sans-serif` scss variable.

**Some questions I still have:**
1. [Bit-supported compilers](https://bit.dev/bit/envs?namespaces=compilers) as a rule do not touch style files (.scss, .css, what have you) since the desired behavior is to let the consuming application process the component's styles. This is fine when we try to render Bit components in Desktop or Bit.dev, but becomes a problem in Cloud since we're not compiling .scss files outside of [this node-sass-middleware stylesheet directory](https://github.com/qri-io/cloud/blob/master/frontend/src/app.ts#L40) (from what I can tell, but please correct if I'm wrong). I am assuming that since we plan to spin up a React app one way or another for the stats diff work that we'll be able to get around this by having webpack available in the context where this Bit component will be rendered to use CSS modules here -- is that a fair assumption to make, @chriswhong? If not, we'll have to get a little more creative for how to compile those files. I haven't played around with touching the internal webpack setups of bit compilers but we could try that out
2. @ramfox you can see [here](https://github.com/qri-io/desktop/compare/au/bit-local-styling?expand=1#diff-fdd4b28079fdfa057e4d44140a1e2fe8R16) what I did to pull scss variables into the component. When you mentioned bringing in these as props for inline styling, is this what you meant? I couldn't think of a better way to pull an scss variable into JS than to have it defined by the consuming app like this, but I'm not a fan of asking developers to remember to do this. I'd rather have the prop be some sort of typescript enum (which is maybe what you were hinting at with the `desktop` or `cloud` modifiers?) but I couldn't figure out a way to make that work without shipping the styles for each enum member with the component (so in the `fontFamily` case, having both the Desktop and Cloud fonts available in the Bit component which is tough since those asset files are really large). Is there a simpler way here to pull in the `$font-family-sans-serif` variable from the consumer app context?


app/components/DetailsBar.tsx
```diff
@@ -29,6 +32,7 @@ const DetailsBarComponent: React.FunctionComponent<DetailsBarProps> = (props) =>
         {renderHeader(statsDetails.title, statsDetails.stats.type)}
         <div className="details-bar-content">
           <StatsChart data={statsDetails.stats} />
+          <StatsChartV1 data={statsDetails.stats} fontFamily={styles.fontFamily}/>
```
uhLeeshUh, line 35: This renders an identical version of the same component immediately above it

app/components/statsChartV1/LabeledStatsV1.tsx
```diff
@@ -0,0 +1,34 @@
...
+  // default is dark
+  color?: 'light' | 'dark'
+  // default sm
+  size?: 'sm' | 'lg'
+  fontFamily: string
+}
+
+const LabeledStats: React.FunctionComponent<LabeledStatsProps> = ({ data, color = 'dark', size = 'sm', fontFamily }) => {
+  return (
+    <div className={styles['labeled-stats']}>
```
uhLeeshUh, line 21: Using SCSS is still fairly new to me so I have a question here -- I was surprised that I had to update the classnames on each of these elements (this one used to just be `'labeled-stats'`) with the css module value. There is a parent `.StatsChartV1` class under which `'labeled-stats'` [is nested](https://github.com/qri-io/desktop/pull/582/files#diff-74c9cd4edb6af347fc304d128c911cecR1-R2) and I [applied that class to the top-level div](https://github.com/qri-io/desktop/pull/582/files#diff-f142f96ab572893976c31d1f9bdfd3b1R70) of each `StatsChart` iteration, but I noticed the general `'label-stats'` styling from `app.global.scss` would overwrite what was nested underneath `.StatsChartV1` so I just updated all of these classnames directly as well. Am I doing something wrong here? I would have expected that just applying the `.StatsChartV1` class to the top-level div would have correctly scoped its children elements

ramfox:
re: scss variables & scss with bit. I think we should actually remove scss from the styling of the components that we are bringing into bit. 
When I was referring to css modules, I was referring to this: https://github.com/css-modules/css-modules/blob/master/docs/css-modules-with-react.md & https://github.com/gajus/babel-plugin-react-css-modules

Sorry if that wasn't clear!

chriswhong:
> I am assuming that since we plan to spin up a React app one way or another for the stats diff work that we'll be able to get around this by having webpack available in the context where this Bit component will be rendered to use CSS modules here -- is that a fair assumption to make, @chriswhong? 

I believe we can just have the express middleware pick up bit scss files along with all the other scss it's currently watching.  Even where there is a small react app running in the client, I think it's reasonable to just have one big CSS file for qri.cloud that includes everything that will ever be in the DOM .

app/components/statsChartV1/LabeledStatsV1.tsx
chriswhong: This isn't a SCSS thing, it's a CSS modules thing.  The styles in `statsChartV1.module.scss` must all be explicitly applied as you have done here.  Remember, the top level `StatsChartV1` gets turned into `_34tMeGqCLyOzIUa0Kbm13Z`, so if you left the `label-stats` in a child  element, the resulting DOM would look like this:

```
<div class="_34tMeGqCLyOzIUa0Kbm13Z">
  <div class="label-stats">
  </div>
</div>
```

The `label-stats` rules defined in the CSS module are never applied because after compilation, it's not `label-stats` anymore, it's `_32_LOBXB03enlW_ucBQMOO`.  This is why the global `label-stats` applied.

All of that said, I think we can have the same pattern without CSS modules, we can just have each bit component import its own scss which will just get wrapped up into the global CSS for the app.

chriswhong:
> Bit-supported compilers as a rule do not touch style files

I'm not clear on what this means as I am able to see the styled bit components on bit.dev.

uhLeeshUh:
> When I was referring to css modules, I was referring to this: https://github.com/css-modules/css-modules/blob/master/docs/css-modules-with-react.md & https://github.com/gajus/babel-plugin-react-css-modules
> 
> not modularizing the scss :)
> 
> Sorry if that wasn't clear!
> 
> I agree that relying on scss might not be good for use with both cloud + desktop!

ahh @ramfox no worries, that was my misunderstanding! Thanks for the links -- I'll do some digging to see which one makes more sense for us

app/components/statsChartV1/LabeledStatsV1.tsx
uhLeeshUh: Ahhh excellent, I see. Thanks for the clarification, that makes sense

uhLeeshUh:
> I believe we can just have the express middleware pick up bit scss files along with all the other scss it's currently watching.

@chriswhong oh that'd be great! I had interpreted that our setup requires all scss files to be co-located in the [same directory](https://github.com/qri-io/cloud/blob/master/frontend/src/app.ts#L40) which meant keeping our Bit component with its own scss file would require somehow importing the Bit scss into this directory. If I'm wrong here though would love to learn how you were imagining the setup

uhLeeshUh:
> > Bit-supported compilers as a rule do not touch style files
> 
> I'm not clear on what this means as I am able to see the styled bit components on bit.dev.

Ahh yes sorry this one is a little confusing. The bit.dev remote as its own "consuming project", as I interpreted from my poking around in [this issue](https://github.com/teambit/bit/issues/1335#issuecomment-452334963-permalink:~:text=All%20you%20will%20have%20to%20do%20is%20to%20name%20your%20scss%20file%20with%20the%20name%20%5Bname%5D.module.scss%20and%20it%20will%20work.), does support .scss files if you suffix them with the `.module` pattern as I did. Other consuming projects however just receive the .scss files untouched and expect you to handle them

chriswhong:
There is an `includePaths` option you can pass in but I think it would require explicit paths to each sass file that needs to be included.  I think another approach is to use another package that allows you to use glob patterns in your sass imports.  https://www.npmjs.com/package/node-sass-glob-importer

uhLeeshUh:
Note: closing this in favor of the node-sass-glob-importer route to be used in Cloud repo

PR #586, opened 2020-08-13
chore(Bit): initialize empty Bit workspace

### LINKED ISSUE #585: chore(Bit): add Bit config to Desktop
In order to support using Bit to package up Desktop components and share them with Cloud, we'll need to add some lightweight config to Desktop. 

The two files we need to ensure our team has the same local working Bit environment are:

1. `bit.json` - bit config file
2. `.bitmap` - file bit uses to keep track of local bit components on your machine

More documentation about both of these files can be found in the Ops repo under the "Bit Configuration" section

uhLeeshUh:
closes #585 

This PR adds the basic config for our team to use Bit by ensuring our local workspaces are configured in the same way.

These two files (`bit.json` and `.bitmap`) are the 2 files that Bit requires us to [commit](https://docs.bit.dev/docs/tutorials/bit-nodejs-tutorial#committing-your-work). `.bitmap` just tracks changes made to local Bit components and `bit.json` is the config for the Bit workspace.


bit.json
```diff
@@ -0,0 +1,23 @@
+{
```
uhLeeshUh, line 1: I included a note about what this file is doing under the "How to configure Bit for our needs?" section in the Bit docs of the Ops repo, but copying it here to make things easier for your review:

Bit configuration is set in the `bit.json` file (or the `package.json`, but we've set it here as a matter of preference).

The basic points to note here are:
 
- We are stating the compiler to use for all Bit components (in this case [`react-typescript`](https://bit.dev/bit/envs/compilers/react-typescript)). If there's ever a situation where you want to use a different compiler for a component, follow [these steps](https://docs.bit.dev/docs/building-components#defining-a-compiler:~:text=You%20can%20also%20use%20the%20overrides,using%20a%20namespace%20or%20glob%20pattern).)
- We are updating the compiler config to target ES5 and utilize CommonJS modules to accomodate older browsers and older Node versions for Cloud server-side rendering
- Any imported components into the project will be located in the `/components` directory
- We've identified a list of `peerDependencies` for packages which we do not want Bit to include in its component's dependency graph. These are packages which we know will be present in the consuming project and can be used as singletons instead of having the Bit component import its own version. See the [Bit documentation](https://docs.bit.dev/docs/react-guidelines#add-react-libraries-as-peer-dependencies-with-relaxed-versions) on this for more info.

b5 (approved):
I think we should merge this to keep the bit project moving!

PR #587, opened 2020-08-27
fix: long dataset history overlaps status button

### LINKED ISSUE #551: Unable to edit dataset when list of commits reaches a certain number
In the QRI desktop, by default the `History` tab on the left is highlighted when editing a local dataset. When the list of commits in this history tab is long enough, I'm unable to click on the `Status` tab to start editing my dataset. The only way to fix this is to maximize the window, at which point I can resume editing. But this too reaches a limit beyond which i can't edit anymore. See screenshot
Screen Shot 2020-05-18 at 11 24 58 AM

chriswhong:
Long sidebar content in the dataset view was preventing the user from clicking the status and history tabs. Closes #551

### DIFF
--- a/app/scss/0.4.0/network.scss
+++ b/app/scss/0.4.0/network.scss
@@ -120,4 +120,6 @@
 
 #history-list {
   padding: 10px;
+  flex-grow: 1;
+  overflow: auto;
 }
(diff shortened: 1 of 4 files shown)


ramfox (approved):
LGTM!

rebased from master and added 1 dep upgrade!

Hope this is okay @chriswhong, but I'm going to merge so I can set up for a release!

PR #588, opened 2020-08-27
refactor(api): adjust `fetchDataset` to use the `/get/` endpoint

ramfox:
We have removed the root endpoint in the backend, so all get requests must go through the `/get` endpoint


ramfox:
ty, now just need to figure out why the tests aren't passing.

Issue #593, opened 2020-09-16
bug(dataset): "show in finder/explorer" in menu shouldn't work if a dataset isn't checked out


ramfox:
[from technickle on discord](https://discord.com/channels/497052800778502152/587775888721051650/755810155307729016):

"Also the "Dataset -> Show in Explorer"  menu command just opens the Qri Desktop folder in %appdata% on my windows machine if you haven't done a checkout yet. So maybe it should bump you to the checkout prompt if you haven't done it first."

We shouldn't allow the "Show in Explorer" menu command if the dataset is not checked out

ramfox:
@dustmop or @b5 , unfortunately I can't repro this on the mac version. It is correctly enabled/disabled when we have a non-checked out dataset or we have a checked out dataset. The logic as written is sound.

We do base whether we enable the "show in explorer" menu on `fsiPath != ''`. Could the windows backend be sending back something different for `fsiPath` then the mac backend, potentially creating a situation where there is content in the `fsiPath` field?

Any chance someone can pull `fix_various_bugs` on a windows machine, add a dataset, and look at the menu for `Dataset => Show in Explorer` before and after checking out the dataset?

dustmop:
I am unable to reproduce this in Windows. When I have a dataset that isn't checked out, the "Show in explorer" option is disabled.

ramfox:
Found an additional bug around this issue. In the menu items that get generated when you right-click a component, it always says 'Open in Finder', need to fix that. Once that is fixed I think we should close this issue as we can't repro the original problem.

PR #598, opened 2020-09-17
fix(*): Fix various bugs

### LINKED ISSUE #552: "Copy Cloud Link" not working on mac
Clipboard does not show the cloud link. 
c.f https://github.com/qri-io/desktop/blob/21c81f2b0c0f0a21df7b0ec4f1d8a5629a362b45/app/components/workbench/headerButtons/PublishButton.tsx#L47

### LINKED ISSUE #591: History list overflows container when viewing a dataset in the network tab
### What is your OS and version?
MacOS Catalina 10.15.5 (19F101)

### What version of qri are you using (`qri version`)?
0.9.12-dev

### What version of Qri Desktop are you using?
0.4.5

## Issue

History list overflows container when viewing a dataset in the network tab.  The panel that contains the list should be scrollable.  Instead the containing element is scrollable, causing a weird display state.

Qri_Desktop

### LINKED ISSUE #592: bug(rename): desktop doesn't allow `-`, but qri core does
Should adjust the validator to allow `-`

https://github.com/qri-io/desktop/blob/981fadc4ba41e38f0fedf4a4470dd8dc0765afa1/app/utils/formValidation.ts#L49

Need to add `-` to this regex:

```
 const invalidCharacters = !(/^(?![0-9])[a-z0-9_-]+$/.test(name))
```

### LINKED ISSUE #595: No padding below commit button
Qri_Desktop

### LINKED ISSUE #596: Pasting text copied from website into description fields results in html
Copying rich text from a website and pasting it into a schema description text input causes the input to contain html.

The workaround has been pasting into a text editor, then copying and pasting again into Qri Desktop.

### LINKED ISSUE #597: Nonfunctional tooltip icons on Commit pane
Qri_Desktop

ramfox:
closes #597 
closes #596 
closes #595 
closes #552 
closes #591 
closes #592

#593, not able to reproduce the error found by a beta user on windows. Asked for some help from @dustmop or @b5 to see if it is a windows only bug

Going to push #593 to a future pr, if we can repro it

### DIFF
--- a/app/components/workbench/components/CommitEditor.tsx
+++ b/app/components/workbench/components/CommitEditor.tsx
@@ -44,9 +44,10 @@ export const CommitEditorComponent: React.FunctionComponent<CommitEditorProps> =
 
   const { username, name } = qriRef
 
-  React.useEffect(() => {
-    ReactTooltip.rebuild()
-  }, [])
+  // The `ReactTooltip` component relies on the `data-for` and `data-tip` attributes
+  // we need to rebuild `ReactTooltip` so that it can recognize the `data-for`
+  // or `data-tip` attributes that are rendered in this component
+  React.useEffect(ReactTooltip.rebuild, [])
 
   React.useEffect(() => {
     // validate form -AND- make sure dataset status is in a commitable state
--- a/app/components/workbench/components/CommitEditor.tsx
+++ b/app/components/workbench/components/CommitEditor.tsx
@@ -3,6 +3,7 @@ import { Action } from 'redux'
 import classNames from 'classnames'
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 import { faSync } from '@fortawesome/free-solid-svg-icons'
+import ReactTooltip from 'react-tooltip'
 
 import Store, { Status, RouteProps } from '../../../models/store'
 import { saveWorkingDatasetAndFetch } from '../../../actions/api'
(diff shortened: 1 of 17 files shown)


ramfox:
@chriswhong it would be great if you could build desktop off this branch and try to repro any of the bugs on this list you filed. The major one would be the "paste into schema description" bug

app/components/workbench/components/CommitEditor.tsx
```diff
@@ -43,6 +44,10 @@ export const CommitEditorComponent: React.FunctionComponent<CommitEditorProps> =
 
   const { username, name } = qriRef
 
+  React.useEffect(() => {
+    ReactTooltip.rebuild()
+  }, [])
```
b5, line 49: Unless this function _must_ be bound to the current context, this should work and be a little cleaner:
```suggestion
  React.useEffect(ReactTooltip.rebuild, [])
```

While we're here can we get a basic comment on what this does? I'm guessing tooltips need to be rebuilt on render, but is there a specific reason? data-model specific stuff in tooltips?

app/components/workbench/headerButtons/PublishButton.tsx
```diff
@@ -32,6 +33,10 @@ const PublishButtonComponent: React.FunctionComponent<PublishButtonProps> = (pro
     setModal
   } = props
 
+  React.useEffect(() => {
+    ReactTooltip.rebuild()
+  }, [])
```
b5, line 38: ditto on the refactoring comment

b5 (approved):
one nit repeated in two places, otherwise LGTM!

chriswhong (approved):
Ran locally and reviewed each bug, everything is looking great!

app/components/workbench/components/CommitEditor.tsx
ramfox: We build the tooltips in the App component. Anything that is not rendered in the first render cycle won't be binded to the ReactTooltip component. So on first render of the component, we rebuild the tooltips.

I'll add the comment!

PR #600, opened 2020-09-28
overhaul app navigation

b5:
Overhaul app navigation to use a hierarchical structure with a back-button-only

Here's an admittedly outdated wireframe:

Screen Shot 2020-09-28 at 9 20 15 AM


ramfox:
I am personally very into removing the status and history tab and replacing them with the above UI, that can be collapsed into just the railroad track of dots (with tooltips that give the commit title).

I have questions about the back button. Is the back button meant as a "last moment in history button", let's say I navigate from version 5 to HEAD, would the back button send me to version 5? Or, is this a back button that sends me to the collection view? I'm assuming the latter, based on the refactor, but want to make sure!

So major changing I'm going to start with:

1) redesign the "sidebar" so that we are only presented with the history track. This sidebar will collapse to just the track, with tooltips that give the commit title, or "working dataset"
2) remove forward button and adjust logic to have the arrow, in the collection view, just return to the collection view
3) move the "dataset action buttons" into the nav, but I'm going to hold off on a redesign for now.

With the correct functionality and navigation set, I think we should merge, and adjust any UI  in a follow-up pr.

chriswhong:
> I am personally very into removing the status and history tab

+1

> Is the back button meant as a "last moment in history button"

I think it should be "back to collection" no matter where you are... it's context doesn't change, if you are anywhere in the dataset view (which will now consist of the dataset editor (status) and the history, it will take you "up" to collection view.

ramfox:
Okay after our checkin, I think the best way forward is to actually merge the nav and router changes, and then open a new pr to deal with the other UI changes I've outlined above. This way, when folks start working on desktop, they will have the correct router in their branch.

PR #610, opened 2020-09-29
feat(Remove dataset): add warning for when a user removes own dataset

uhLeeshUh:
Closes #534 

Here's a preview of what the modal now looks like when a user attempts to delete a dataset under their own username
Screen Shot 2020-09-29 at 12 00 26 PM

PR #611, opened 2020-09-29
Update "Save Path" Checkout modal copy 

### LINKED ISSUE #205: should validate that "Save Path" is not a subset of "Data File"'s path
We should make sure we correctly message, and also enforce, that the Save Path is a separate location from where the "Data file" is being imported from. If the user is importing "~/datasets/new_ds/my_data.json", we should not allow the save path "~/datasets/new_ds/". This should be enforced on the backend as well, but the New Dataset dialog also needs to explain the meaning of these fields.

uhLeeshUh:
Closes #205 

Update copy from "Save Path" to "Checkout Location" and include help text to clarify what the fields in the checkout modal refer to

Screen Shot 2020-09-29 at 1 36 36 PM

### DIFF
--- a/app/components/modals/LinkDataset.tsx
+++ b/app/components/modals/LinkDataset.tsx
@@ -142,7 +142,7 @@ export const LinkDatasetComponent: React.FunctionComponent<LinkDatasetProps> = (
           <TextInput
             name='datasetDirectory'
             label='Dataset Directory'
-            helpText="The name of the directory which contains this dataset's components"
+            helpText="The name of the directory which will contain this dataset's components"
             showHelpText
             type=''
             value={datasetDirectory}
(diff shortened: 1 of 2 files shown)


uhLeeshUh:
just as a note - I pushed a fix to #610  which solves the failing build of e2e tests

b5 (approved):
Looks great. My only gripe: Changes this small ideally don't have _so_ many commits. If you have the time I'd appreciate rebasing this change into a single commit before merging, but not required

uhLeeshUh:
>Looks great. My only gripe: Changes this small ideally don't have so many commits. If you have the time I'd appreciate rebasing this change into a single commit before merging, but not required

@b5 For sure! I normally use the squash and merge approach with PRs and presumed I'd do that here as well. Does the team prefer to rebase in general and keep all commits? Happy to follow whatever pattern we have

PR #620, opened 2020-10-01
fix(*): clean up `discardChanges`, double-fetching bugs, and the "uncommited changes" prompt

### LINKED ISSUE #613: bug(`EditDataset`): 'discard changes' does nothing
if the dataset is checked out, 'discard changes' (right clicking on the component in EditDataset mode) goes thorough the actions to discard changes (`/restore`), but nothing is reflected to the user. aka, the changes are still being rendered.

if the dataset is not checked out, 'discard changes' erroneously sends actions to the backend, as if the dataset is checkedout.

Need to check the reducers & any logic that is specifically about 'status' and 'modified dataset'

### LINKED ISSUE #617: bug(collection): datasets with no history have "updated" time at 2020 years ago
Looks like we may have a zero value datatime getting passed in somewhere. When a dataset has no last commit, we should be showing the `--` text like we are in the other fields
Screen Shot 2020-09-30 at 11 43 24 AM

ramfox:
closes #617 
closes #613 

A bunch of this cleans up functionality that was changed when we refactored the routes and refactored away from the "status" and "history" tabs.

- add notes on how to integrate new modals
- fixed bug in `workbenchRoutes` that was causing the history ref to reset when there was a new edit ref, and the edit ref to reset when there was a new history ref
- fixed bug that was causing double fetching b/c we weren't storing `peername` and `name` in the state tree when we fetched any dataset information
- refactored the "you have uncommitted changes" code into one component for easier future fixes & refactored to ensure it was working as expected
- restored the ability to `Discard changes` in linked and unlinked datasets.

### DIFF
--- a/app/components/collection/UncommitedChangesPrompt.tsx
+++ b/app/components/collection/UncommitedChangesPrompt.tsx
@@ -1,0 +1,61 @@
+import React from 'react'
+import { Prompt, RouteProps } from 'react-router-dom'
+import { Action } from 'redux'
+
+import { connectComponentToPropsWithRouter } from '../../utils/connectComponentToProps'
+import { QriRef } from '../../models/qriRef'
+import { resetMutationsDataset, resetMutationsStatus } from '../../actions/mutations'
+import { selectMutationsIsDirty, selectRecentEditRef } from '../../selections'
+
+interface UncommitedChangesPromptProps extends RouteProps {
+  recentEditRef: QriRef
+  modified: boolean
+  resetMutationsDataset: () => Action
+  resetMutationsStatus: () => Action
+}
+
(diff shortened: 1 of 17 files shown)


app/components/collection/UncommitedChangesPrompt.tsx
```diff
@@ -0,0 +1,61 @@
+import React from 'react'
```
chriswhong, line 1: When testing this locally, I noticed that it only fires when moving to another top-level route.  

I created a new dataset, edited readme, then clicked on a commit in the history list... when I click back on working dataset, my changes are gone.

chriswhong (changes requested):
Tested restoring changes locally, everything looks great.  Only issue is the uncommitted changes warning not firing when moving from working dataset to a history item

b5 (approved):
RUBBER STAMP

PR #622, opened 2020-10-01
feat: export modal and css improvements

chriswhong:
Implements an advanced export modal.
- User can choose body and then choose between CSV or JSON
- User can choose meta, structure, and readme to include in the export
- User can select/deselect all "other components" with a checkbox
Qri_Desktop

## Other general improvements

- Adds a new component `CommitDetails` which is used in the above Modal, and also used in the header of the dashboard (standardizing the style, icons, layout, etc.)

- Replaces the disk icon used for "Commit" with the line and circle commit icon from cloud.

- Improves whitespace in collection view

- Eliminates double vertical scrollbar when viewing dataset bodies in the workbench

- Styles scrollbars to make them thinner and less intrusive

- Removed sidebar header over the dataset history (showing redundant info)
Fullscreen_10_1_20__11_59_AM


app/components/CommitDetails.tsx
```diff
@@ -0,0 +1,30 @@
...
+  structure: Structure
+  commit: Commit
+  path: string
+}
+
+const CommitDetails: React.FunctionComponent<CommitDetailsProps> = (props) => {
+  const { structure, commit, path } = props
+  return (
+    <>
+      <span className='dataset-details' style={{ marginBottom: '6px' }}><Icon icon='stickyNote' size='sm' color='default' />{commit.title}</span>
```
uhLeeshUh, line 19: just out of curiosity, when do we typically use inline style tags vs scss?

app/components/RelativeTimestamp.tsx
```diff
@@ -0,0 +1,17 @@
...
+import moment from 'moment'
+
+interface RelativeTimestampProps {
+  timestamp: string
+}
+
+const RelativeTimestamp: React.FunctionComponent<RelativeTimestampProps> = ({ timestamp }) => (
+  <span
+    className='relative-timestamp'
+    title={moment(timestamp).format('MMM D YYYY, h:mm A zz')}
```
uhLeeshUh, line 11: such a nice touch

app/components/collection/layouts/DatasetMainContent.tsx
```diff
@@ -19,28 +15,12 @@ interface DatasetMainContentProps extends RouteProps {
...
-          return true
-        }
-        return `You have uncommited changes! Click 'cancel' and commit these changes before you navigate away or you will lose your work.`
-      }}/>
-      <LinkButton />
-      <PublishButton />
-    </div>
-    {children}
-  </>
+  const { children } = props
```
uhLeeshUh, line 18: Do we need this component anymore if it looks like none of its props are used?

app/components/collection/layouts/DatasetSidebar.tsx
```diff
@@ -30,10 +25,6 @@ export const DatasetSidebarComponent: React.FunctionComponent<DatasetSidebarProp
 
   return (
     <div className='sidebar'>
-      <div className='sidebar-header sidebar-padded-container'>
-        <DatasetReference qriRef={qriRef} />
```
uhLeeshUh, line 34: Don't know if this is an unpopular opinion, but I think removing the dataset ref here takes away an important visual for the user of what they're working on. In your mock, the dataset ref title is now smaller and gray at the top which could get overlooked. What do you think of removing the redundant details subtext but keeping this ref?

app/components/form/CheckboxInput.tsx
```diff
@@ -1,25 +1,35 @@
...
           name={name}
           type='checkbox'
           className='checkbox-input'
           checked={checked}
           onChange={ () => { onChange(name, !checked) }}
+          ref={el => el && (el.indeterminate = indeterminate)}
+          disabled={disabled}
         />
-        {label}
+        { strong && (
```
uhLeeshUh, line 29: Could likely turn these two checks into one with a ternary if the only two states are `strong` and `!strong`

app/components/form/CheckboxInput.tsx
```diff
@@ -1,25 +1,35 @@
...
   onChange: (name: string, checked: any) => void
+  strong?: boolean
+  disabled?: boolean
 }
 
-const CheckboxInput: React.FunctionComponent<CheckboxInputProps> = ({ label, name, checked, onChange }) => {
+const CheckboxInput: React.FunctionComponent<CheckboxInputProps> = (props) => {
+  const { label, name, checked, onChange, indeterminate = false, strong = false, disabled = false } = props
   return (
     <>
```
uhLeeshUh, line 17: small nit here and I know it's not code you added, but looks like if this div is the parent element of everything else, could likely remove this React fragment?

app/components/modals/ExportDataset.tsx
```diff
@@ -0,0 +1,220 @@
...
+
+  const [dismissable, setDismissable] = React.useState(true)
+
+  const [error, setError] = React.useState('')
+  const [loading, setLoading] = React.useState(false)
+
+  const [ bodyChecked, setBodyChecked ] = React.useState(true)
+  const [ bodyFormat, setBodyFormat ] = React.useState('csv')
+
+  // otherChecked has three states, 'true', 'false', 'indeterminate'
```
uhLeeshUh, line 45: What does it mean for `otherChecked` to be indeterminate? I'm having trouble following the logic for when it would be possible to check more options than are enabled

app/components/modals/ExportDataset.tsx
```diff
@@ -0,0 +1,220 @@
...
+      setOtherChecked('true')
+    } else {
+      setOtherChecked('indeterminate')
+    }
+  }, [metaChecked, structureChecked, readmeChecked])
+
+  const handleSubmit = () => {
+    setDismissable(false)
+    setLoading(true)
+    error && setError('')
```
uhLeeshUh, line 78: Should this error logic be reversed? I'm thinking if `error` is true once the `onSubmit` is called then you `setError` to that, otherwise `setError` back to empty string?

app/components/modals/ExportDataset.tsx
```diff
@@ -0,0 +1,220 @@
...
+  const [ structureChecked, setStructureChecked ] = React.useState(false)
+  const [ readmeChecked, setReadmeChecked ] = React.useState(false)
+
+  const enabledOtherComponents = ['structure'] // structure is always enabled
+  if (meta) enabledOtherComponents.push('meta')
+  if (readme) enabledOtherComponents.push('readme')
+
+  // whenever a child checkbox is changed, determine the state of the parent
+  // checkbox by comparing count of checked boxes to the count of enabled boxes
+  React.useEffect(() => {
```
uhLeeshUh, line 58: I'm wondering now if we actually need the `otherChecked` field in state? This question is pending my understanding of its indeterminate state though, so feel free to ignore if that's the lynchpin here -- but I'm thinking that we could render the parent "Other Components" radio button as checked if `metaChecked || structureChecked || readmeChecked`, right? And by doing that we avoid the re-render that would happen via this useEffect hook (right now a state change happens for either `metaChecked`, `structureChecked` or `readmeChecked`, the component re-renders, then this hook is called, then we set `otherChecked` which causes another render)

app/scss/_site.scss
```diff
@@ -804,3 +804,21 @@ b {
     opacity: 1
   }
 }
+
+// webkit scrollbars
+/* Let's get this party started */
```
uhLeeshUh, line 809: 🎊

app/components/CommitDetails.tsx
```diff
@@ -0,0 +1,30 @@
+import * as React from 'react'
+import numeral from 'numeral'
+
+import { Structure, Commit } from '../models/dataset'
+
+import RelativeTimestamp from './RelativeTimestamp'
+import Icon from './chrome/Icon'
```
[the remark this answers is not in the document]
chriswhong: Done!

app/components/CommitDetails.tsx
chriswhong: This is a special nudge for this line that isn't repeated anywhere and I don't want to make a class and add CSS rules for it.

app/components/RelativeTimestamp.tsx
chriswhong: TY

app/components/collection/layouts/DatasetMainContent.tsx
chriswhong: I'd leave it for now, I have a feeling something else will end up in this div.

stories/Collection.stories.tsx
```diff
@@ -41,21 +39,12 @@ const datasetStatusBaseProps = {
...
 
 const datasetPublishedProps: DatasetStatusProps = {
-  ...datasetStatusBaseProps,  
+  ...datasetStatusBaseProps,
   published: true,
   linkedToFilesystem: false,
   updatesAvailable: false,
 }
 
-export const datasetPublished = () => tooltipWrapper(<DatasetStatus {...datasetPublishedProps} />, 'Dataset Published')
```
uhLeeshUh, line 50: We're still using this dataset published glyph, right? Why the removal here?

app/components/collection/layouts/DatasetSidebar.tsx
chriswhong: The purpose of not having the dataset ref in this section is that it's not wide enough to display dataset refs that can be hundreds of characters long.  The purpose of a full-width header was so we have have as much space as possible to show the dataset reference.

app/components/form/CheckboxInput.tsx
chriswhong: Done

app/components/form/CheckboxInput.tsx
chriswhong: Good catch, this was leftover from when there was another element in here.

app/components/modals/ExportDataset.tsx
chriswhong: Indeterminate state is when the checkbox shows a "-" to indicate that only a portion of it's subordinate checkboxes are checked.  Thus, it has three states in memory so I use strings instead of a boolean.

app/components/modals/ExportDataset.tsx
chriswhong: I think the idea is to clear out the existing error before firing an API call.  This will remove it from the screen while the spinner is spinning, if a new error comes back, set it, otherwise the call was successful.

uhLeeshUh: ahh I see, so we'll have another bit of error handling after the API call when that's wired up. Sounds good

app/components/modals/ExportDataset.tsx
uhLeeshUh: Gotcha. Could you update this comment with that gist so it's a bit clearer?

app/components/modals/ExportDataset.tsx
chriswhong: I think this is possible in theory, but we would need to modify `CheckboxInput` to return the indeterminate state `onChange` (Right now it only does onChange(name, !checked) ). Since this is the only indeterminate checkbox in the app, I'd rather keep track of it with component state.

uhLeeshUh: Logic here seems complex enough that keeping this as-is is fine with me then

app/scss/_site.scss
chriswhong: Lol, that was copypasta, will remove

uhLeeshUh (approved):
Just a note about updating the comment for what indeterminate means and a question about the storybook story you took out, but otherwise LGTM!

app/scss/_site.scss
uhLeeshUh: I loved it haha

app/components/modals/ExportDataset.tsx
```diff
@@ -0,0 +1,221 @@
+import * as React from 'react'
```
b5, line 1:
```tsx
import React, { useState, useEffect } from 'react'
```

app/components/modals/ExportDataset.tsx
```diff
@@ -0,0 +1,221 @@
...
+  const [ bodyChecked, setBodyChecked ] = React.useState(true)
+  const [ bodyFormat, setBodyFormat ] = React.useState('csv')
+
+  // otherChecked has three states, 'true', 'false', 'indeterminate'
+  // indeterminate state means the checkbox shows a '-' instead of checked or unchecked
+  const [ otherChecked, setOtherChecked ] = React.useState('false')
+
+  const [ metaChecked, setMetaChecked ] = React.useState(false)
+  const [ structureChecked, setStructureChecked ] = React.useState(false)
+  const [ readmeChecked, setReadmeChecked ] = React.useState(false)
```
b5, line 51: group states to force them all to be set at once. wrap the setter in a function that performs the business logic happening in `useEffect`:

```tsx
const [ checked, setChecked ] = useState({
  other: 'false',
  meta: false,
  structure: false,
  readme: false
})

const updateCheckedField = (field: string, value: boolean) => {
  const update = { ...checked, [field] : value }
  // ... logic that affirms valid state
  setChecked(update)
}
```

Note that since users can't "set" the indeterminate field, the accepted `value` argument needs only be boolean.

app/components/nav/NavTopbar.tsx
```diff
@@ -60,6 +65,7 @@ export const NavTopbar: React.FunctionComponent<NavTopbarProps> = ({ title, loca
                 return <HeaderColumnButton key={props.id} {...props} />
             }
           })}
+          <div onClick={handleExportButtonClick}>export</div>
```
b5, line 68: this shouldn't have made it into master as-is

b5 (commented):
I was rebasing on this commit and found a few things I think could be better, I've added notes for future reference. @uhLeeshUh & @chriswhong, plz have a read through when you get a chance. I'll incorporate the fixes in a follow-up

PR #623, opened 2020-10-01
refactor(export dataset): move export config logic to ipcMain handler

### LINKED ISSUE #538: Export creates a zip in a directory
In v0.4.1, choosing "Export this version":

The OS file picker opens, I specify foo.zip on my Desktop
After clicking save, a directory called foo.zip is added to my Desktop.
Inside /foo.zip is a zip file with an autogenerated name: nyc-transit-data-turnstile_daily_counts_2019-04-04-2020-08-04-24-am.zip
What I expected to happen:

Desktop should create foo.zip, a zip file on my desktop, containing the dataset's files.

uhLeeshUh:
Closes #538 

Making this a draft now pending changes we make to the export process. After fixing the export URL I can't reproduce the original error, so planning to just close out that ticket with this refactor since they're related but the orig problem is no longer an issue

### DIFF
--- a/app/components/collection/LogList.tsx
+++ b/app/components/collection/LogList.tsx
@@ -57,8 +57,8 @@ export const LogListComponent: React.FunctionComponent<LogListProps> = (props) =
 
     const directory = path.dirname(selectedPath)
     const filename = path.basename(selectedPath)
-    const exportUrl = `http://localhost:2503/get/${refStringFromQriRef(qriRef)}?format=zip`
-    ipcRenderer.send('export', { url: exportUrl, filename: filename, directory: directory })
+    const refString = refStringFromQriRef(qriRef)
+    ipcRenderer.send('export', { refString, filename: filename, directory: directory })
   }
 
   return (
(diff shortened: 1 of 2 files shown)


uhLeeshUh:
@chriswhong this bug as was reported no longer exists (I couldn't repro), but I am wondering if there's still an issue with exporting in that no save dialog pops up for confirmation of the success or failure of an export event. Am I interpreting [this line ](https://github.com/qri-io/desktop/blob/master/app/components/collection/LogList.tsx#L52)of code correctly as supposed to be triggering some sort of save dialog?

chriswhong:
> Or is that just opening up the native file picker

Yes, this is where the user chooses where to save the zip using the OS file picker.

uhLeeshUh:
@ramfox I found this zip bug actually doesn't exist, so this PR is just making the change we talked about where the export logic that constructs the exportUrl all happens now in the ipcMain handler (which will make it easier for us to refactor as we think about advanced export work)

ramfox (approved):
Dope! LGTM

I think this actually may have been a backend bug that got fixed when dustmop refactored the zip functionality into the `get` endpoint/command.

PR #625, opened 2020-10-01
transfers UI

b5:


b5 (commented):
Ok, performed the best self-review I could muster. Merging!

Issue #496, opened 2020-03-25
Typing in commit title/message fields always appends at end


hbruch:
Trying to add some text in the middle of already entered text in the commit title/message fields will transfer the cursor to the end of the already entered text after the first char is typed.

rustprooflabs:
I encountered this bug today when trying to make minor updates to a long commit message.  I placed the cursor in the middle of the text where edits were needed.  The first typed character is in the appropriate place then the cursor jumps to the very end of the message for the second character and beyond.

Desktop version 0.4.2.

Mr0grog:
I also hit this while creating a new dataset today, and was briefly confused by the gibberish in my commit message when I finished typing. 😜

PR #626, opened 2020-10-01
fix(commit): Don't push cursor to end of text input elements

dustmop:
According to: https://stackoverflow.com/questions/28922275/in-reactjs-why-does-setstate-behave-differently-when-called-synchronously/28922465#28922465, using `onChange` along with `value` on input elements leads to problems when the element is changed asynchronously. Using `defaultValue` instead fixes this problem.

Fixes https://github.com/qri-io/desktop/issues/496.


b5 (approved):
fantastic. this is a _very_ welcome 2 line change.

PR #627, opened 2020-10-01
refactor(TopNavbar): remove ForwardArrow & update BackArrow to return to base route

### LINKED ISSUE #607: refactor(NavTopbar): remove forward arrow & back arrow should take you to the base url of that view
Eg, if we are on the collections page & then click into a dataset, the "back" arrow should return us to the collections page.

If we are in the network tab & we click into a dataset, the "back" arrow should return us to the network page.

uhLeeshUh:
closes #607 

- removes ForwardArrow component completely
- updates BackArrow logic to always return to base route. If at base route, then arrow is disabled


ramfox (approved):
LGTM!!!!

app/components/nav/NavTopbar.tsx
```diff
@@ -42,12 +41,21 @@ export const NavTopbar: React.FunctionComponent<NavTopbarProps> = ({ title, loca
...
+    return history.push(`/${baseRoute}`)
+  }
+
   return (
     <div className='page-navbar'>
       <div className='row'>
         <div className='nav-buttons'>
-          <a className='back' onClick={() => history.goBack()}><BackArrow /></a>
-          <a className='forward' onClick={() => history.goForward()}><ForwardArrow /></a>
+          {!isBaseRoute && <a className='back' onClick={onBackClick}><BackArrow /></a>}
```
b5, line 58: I've also adjusted the back arrow to not appear at all whenever we're at a base route. keeping a convention established with mobile apps.

b5 (approved):
made some quick adjustments. LGTM!

app/components/nav/NavTopbar.tsx
uhLeeshUh: sounds great, thanks!

app/components/nav/NavTopbar.tsx
```diff
@@ -42,12 +41,21 @@ export const NavTopbar: React.FunctionComponent<NavTopbarProps> = ({ title, loca
     setModal({ type: ModalType.ExportDataset })
   }
 
+  // determines if route is at base route (e.g. /collection or /network)
+  const isBaseRoute = Object.keys(match.params).length === 0
```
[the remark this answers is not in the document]
uhLeeshUh: No prob. Sounds good to me

PR #628, opened 2020-10-02
fix(dataset): add back publish & show files buttons

b5:
I'd prefer to see these two buttons refactored, but this is the faster way to get them back


ramfox (approved):
LGTM!

PR #629, opened 2020-10-02
fix: external readme links

### LINKED ISSUE #540: No way back from links in readme / Readme links to open in default browser
From a user:

"If you click on the link in the readme, it opens the address in the Qri app with no way to close out or back out unless you quit the app. (It's pretty frustrating when you first encounter). It looks like command+click (I'm on a Mac) open the link in its own little window inside the app but Ideally, I would want those links to just open on my default browser which where it is a much better experience to navigate."

chriswhong:
Adds click listeners to hijack clicks on rendered readmes (both in historic commit readme display and when previewing a working readme), so that links are intercepted and sent to electron's external link opener.

Includes some css cleanup to fix a double vertical scrollbar on the readme editor and a non-scrolling structure viewer.  All components should overflow/scroll in the same way now.

Removes `color='light'` prop on WorkingComponentList items so each component's icon is shown (they were white on white and not visible)

closes #540

### DIFF
--- a/app/components/collection/datasetComponents/Readme.tsx
+++ b/app/components/collection/datasetComponents/Readme.tsx
@@ -2,11 +2,16 @@ import * as React from 'react'
 
 import Store, { RouteProps } from '../../../models/store'
 import { refStringFromQriRef, QriRef, qriRefFromRoute } from '../../../models/qriRef'
+import { onClick as openExternal } from '../../platformSpecific/ExternalLink.electron'
 
 import { connectComponentToProps } from '../../../utils/connectComponentToProps'
 
 export interface ReadmeProps extends RouteProps {
   qriRef: QriRef
+}
+
+const handleClick = (e: React.SyntheticEvent) => {
+  openExternal(e, e.target.href)
 }
 
 export const ReadmeComponent: React.FunctionComponent<ReadmeProps> = (props) => {
@@ -40,6 +45,7 @@ export const ReadmeComponent: React.FunctionComponent<ReadmeProps> = (props) => 
       // use "editor-preview" class to piggie-back off the simplemde styling
       className="editor-preview"
       ref={ref}
+      onClick={handleClick}
     >loading...
     </div>
   )
--- a/app/components/collection/datasetComponents/Readme.tsx
+++ b/app/components/collection/datasetComponents/Readme.tsx
@@ -2,7 +2,7 @@ import * as React from 'react'
 
 import Store, { RouteProps } from '../../../models/store'
 import { refStringFromQriRef, QriRef, qriRefFromRoute } from '../../../models/qriRef'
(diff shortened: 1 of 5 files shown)


app/components/collection/datasetComponents/Readme.tsx
```diff
@@ -2,13 +2,18 @@ import * as React from 'react'
...
 import { refStringFromQriRef, QriRef, qriRefFromRoute } from '../../../models/qriRef'
+import { onClick as openExternal } from '../../platformSpecific/ExternalLink.electron'
 
 import { connectComponentToProps } from '../../../utils/connectComponentToProps'
 
 export interface ReadmeProps extends RouteProps {
   qriRef: QriRef
 }
 
+const handleClick = (e: React.SyntheticEvent) => {
```
b5, line 13:
```suggestion
const handleClick = (e: React.MouseEvent) => {
```

app/components/collection/datasetComponents/ReadmeEditor.tsx
```diff
@@ -8,6 +8,7 @@ import hasParseError from '../../../utils/hasParseError'
 import Dataset from '../../../models/dataset'
 import { refStringFromQriRef, QriRef, qriRefFromRoute } from '../../../models/qriRef'
 import Store, { StatusInfo, RouteProps } from '../../../models/store'
+import { onClick as openExternal } from '../../platformSpecific/ExternalLink.electron'
```
b5, line 11:
```suggestion
import { onClick as openExternal } from '../../platformSpecific/ExternalLink.TARGET_PLATFORM'
```

app/components/collection/datasetComponents/ReadmeEditor.tsx
```diff
@@ -85,6 +90,11 @@ export const ReadmeEditorComponent: React.FunctionComponent<ReadmeEditorProps> =
    * are okay living where they work
    */
   const getPreview = (plainText: string, preview: HTMLElement) => {
+    if (!listenerAdded) {
+      preview.addEventListener('click', passEventToOpenExternal)
+      setListenerAdded(true)
+    }
```
b5, line 96: this seems a like a sketchy use of state, but meh, I can't think of a quicker way to fix this

b5 (approved):
use `TARGET_PLATFORM` instead of `electron`, looks good otherwise

app/components/collection/datasetComponents/ReadmeEditor.tsx
chriswhong: Agree, but it works and seems to be stable.  The div we need to listen for clicks on doesn't exist until the user chooses to preview their markdown.  AFAIK this is the first place we have access to it without hacking the library.

PR #630, opened 2020-10-02
test(e2e): get e2e tests to pass with refactorings

ramfox:
**There are two tests that are still failing, both in the same way.** When we go to switch between commits & check that the commit titles are correct (this is how we determine that the versions have actually been changed, since we don't have the path information and therefore can't use the path to verify we have navigated to the correct commit). For some reason, even though the div definitely exists, we cannot find the `#history-commit` on the page. Would love it if someone could take a look.

However, this is in a good enough place that even if someone isn't able to de-bug, I'm comfortable merging, since, for most cases, folks can use this to make sure they aren't breaking expected app behavior. 

Three notes:
1) DatasetReference was added back in because otherwise there is no way to rename a dataset in the app
2) the LinkDataset and PublishDataset components were added back in because otherwise there is no way to checkout or publish/unpublish a dataset in the app
3) I added back the `CreateDataset` component (which was replaced by the `NewDataset` component), because the flow from the `NewDataset` modal, which doesn't have space to add a body file, is not finished. This flow needs to be finished before we add the `NewDataset` modal back in, because otherwise the desktop app cannot add datasets.

Otherwise, all the other tests are passing locally!

### DIFF
--- a/app/components/collection/LogList.tsx
+++ b/app/components/collection/LogList.tsx
@@ -16,7 +16,7 @@ import { selectLog, selectLogPageInfo, selectInNamespace, selectRecentEditRef } 
 
 import LogListItem from '../item/LogListItem'
 import { pathToDataset, pathToEdit } from '../../paths'
-import HeadLogListItem from '../item/HeadLogListItem'
+import WorkingLogListItem from '../item/WorkingLogListItem'
 
 interface LogListProps extends RouteProps {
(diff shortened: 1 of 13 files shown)


b5 (approved):
Ok, this looks great. I've rebased & run this against a rebased branch. The best path forward is to merge this & iterate. LGTM!

PR #632, opened 2020-10-02
feat(collection home): support bulk pulling

b5:
add our first bulk action. A few things left for cleanup here:
* buttons should dim while bulk action is executing. they're disabled during the whole time
* the toast that shows 'pulling X datasets' should  stay up, and not have a warning
* currently error responses from the backend are still reporting success?


b5:
ok, I'm going to merge this so any conflicts with #633 can be incorporated when adding in review feedback

PR #633, opened 2020-10-03
Responsive & Sortable Collection Table

chriswhong:
Screen Shot 2020-10-02 at 11 34 54 PM

Refactor of Collection view table to use [`react-data-table-component`](https://github.com/jbetancur/react-data-table-component), allowing for flex columns, sorting by name, updated, size, rowcount, etc.

`VersionInfoItem` is removed, all of the display logic for the different columns is now defined in `DataTable`.

- Adds a hamburger menu to each row with `Remove` action, can be used to add more dataset-specific actions as needed
- Cleans up layout of the app header, ensuring a fixed height and no jumping between views
- Adds a two-line layout for dataset reference in the header in the workbench view


ramfox:
Haven't reviewed this, but wanted to point out: we rely on the `${username}-${datasetname}-link` html id tag in `VersionInfoItem` to navigate through the app in the e2e tests, please make sure that there is a replacement in the data table on whatever element we click on to get to the dataset page

https://github.com/qri-io/desktop/blob/8bf137a4faedb4d581a9c6d39812b75d8bb90ff4/app/components/item/VersionInfoItem.tsx#L43

chriswhong:
Our data table library will use the `id` property in each object that becomes a row to make an `id` on the element, but it prepends `row-`.  

I just pushed a change that adds `id` to each object in `filteredDatasets` for this purpose, and updated the e2e tests to use `row-${username}-${name}` instead of ``${username}-${name}-link`

ramfox:
you rock thank you!

app/components/collection/collectionHome/StatusIcons.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react'
+
+import { VersionInfo } from '../../../models/store'
+
+import ExternalLink from '../../ExternalLink'
+import Icon from '../../chrome/Icon'
+import { QRI_CLOUD_URL } from '../../../constants'
+
+interface StatusIconsProps {
+  row: VersionInfo
```
b5, line 10: I'd prefer this be named `data`

app/components/nav/NavTopbar.tsx
```diff
@@ -1,18 +1,18 @@
 import React from 'react'
 import { IconDefinition } from '@fortawesome/free-regular-svg-icons'
+import classNames from 'classnames'
 
 import { RouteProps } from '../../models/Store'
 import { connectComponentToPropsWithRouter } from '../../utils/connectComponentToProps'
 
 import BackArrow from '../chrome/BackArrow'
-import Breadcrumb from '../chrome/Breadcrumb'
```
b5, line 8: Is breadcrumb totally gone now? If so delete the component

app/components/nav/NavTopbar.tsx
```diff
@@ -1,18 +1,18 @@
...
 import HeaderColumnButton from '../chrome/HeaderColumnButton'
 import HeaderColumnButtonDropdown from '../chrome/HeaderColumnButtonDropdown'
 import Transfers from '../Transfers'
 
 // RouteProps includes `history`, `location`, and `match`
 interface NavTopbarProps extends RouteProps {
-  title: string
+  title: string | React.ReactElement
```
b5, line 15: passing an element here is overkill. Instead, accept an optional `subtitle` prop & inline all styling & markup into this component.

b5 (commented):
Looking good! Some required fixes

PR #638, opened 2020-10-06
fix(pull): adjust flow after a dataset is pulled

### LINKED ISSUE #635: bug(pull): after pulling a dataset, you should be taken to the most recent version of the dataset
"When pulling another user’s dataset, user lands on workbench with no username/name in the header, and “You are viewing another user’s dataset” message. Goes away after reload or navigate to collection and back (It looks like the fix for this is to select the HEAD commit immediately after pulling)"

ramfox:
Also removes dead code and adjusts e2e tests

closes #635


app/components/collection/Collection.tsx
```diff
@@ -4,34 +4,27 @@ import { ipcRenderer, shell } from 'electron'
...
 
 import {
   selectFsiPath,
-  selectDetails,
   selectWorkingDatasetIsPublished,
   selectSessionUsername,
   selectShowDetailsBar,
   selectSidebarWidth,
-  selectHasDatasets
+  selectHasDatasets, selectLogLatestVersion
```
b5, line 19: add newline

app/selections.ts
```diff
@@ -116,6 +116,10 @@ function datasetFromDatasetStore (historyDataset: DatasetStore): Dataset {
  *
  */
 
+export function selectLogLatestVersion (state: Store): VersionInfo | undefined {
+  return selectHasHistory(state) ? selectLog(state)[0] : undefined
+}
```
b5, line 121: don't  use the ternary operator to return undefined

```suggestion
export function selectLogLatestVersion (state: Store): VersionInfo? {
  if (selectHasHistory(state)) {
    return selectLog(state)[0]
  }
}
```

also, does the optional typescript operator work on function signatures? I'd love to use it as written  above if so

b5 (approved):
2 nits. otherwise,  delighted to see all this cleanup!

app/selections.ts
ramfox: i wonder how many times you are going to tell me not to use the ternary operator w/ undefined before i internalize it 🤦‍♀️

ramfox: also sadly no ? for optional returns, but if you want to read about the discussion surrounding it: https://github.com/microsoft/TypeScript/issues/7426

PR #640, opened 2020-10-07
fix(body): relax stipulations for fetching next page

### LINKED ISSUE #639: bug(Body): body pagination is not working
When you scroll to the bottom of the body, and there are more rows in the body than are currently visible, the app should fetch the next batch of rows

ramfox:
Previously, we were only fetching the next page if the scrollHeight was
exactly equal to the scrollTop + the offsetHeight. It looks like this
calculation was thrown off by style changes and the scroll-x bar. We now
include a 'buffer' into the calculation. If the scrollHeight is within
10px of the scrollTop + offsetHeight, we attempt to fetch (barring that
we have already `fetchedAll`)

closes #639

### DIFF
--- a/app/components/BodyTable.tsx
+++ b/app/components/BodyTable.tsx
@@ -34,6 +34,10 @@ const getNextPageNumber = (direction: 'up' | 'down', value: any[], pageSize: num
   return firstPage - 1
 }
 
+const shouldFetchNextPageDown = (scrollHeight: number, scrollTop: number, offsetHeight: number, buffer: number) => {
+  return Math.abs(scrollTop + offsetHeight - scrollHeight) < buffer
+}
+
 export default class BodyTable extends React.Component<BodyTableProps> {
   constructor (props: BodyTableProps) {
     super(props)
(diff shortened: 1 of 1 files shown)


app/components/BodyTable.tsx
```diff
@@ -34,6 +34,10 @@ const getNextPageNumber = (direction: 'up' | 'down', value: any[], pageSize: num
   return firstPage - 1
 }
 
+const shouldFetchNextPageDown = (scrollHeight: number, scrollTop: number, offsetHeight: number, buffer: number): boolean => {
+  return Math.abs(scrollTop + offsetHeight - scrollHeight) < buffer
```
uhLeeshUh, line 38: Ah so we'd expect scrollTop + offsetHeight to equal scrollHeight, right? But we're giving it a bit of a buffer here just in case that's not so. Nice!

Also the smallest of nits here, but I'm pretty sure TS can infer the boolean return type of the fn? So you can likely remove the `: boolean`

uhLeeshUh (approved):
Nice! Tried it out and looks 💯

PR #643, opened 2020-10-08
feat(export): wire up export modal with csv & zip options

b5:
we don't have support for per-component export in the backend at the moment, so I've moved the dataset component picking stuff into it's own react component for future use in an "advanced" export section.

I'd also like the option to export the entire dataset as a JSON file. I use that format a bunch, it'd be great to have at some point in the future.

### DIFF
--- a/app/components/form/ComponentAndFormatPicker.tsx
+++ b/app/components/form/ComponentAndFormatPicker.tsx
@@ -15,6 +15,8 @@ interface ComponentAndFormatPickerProps {
 interface ComponentConfig {
   body: boolean
 
+  // tri-state enum of all following fields
+  // indeterminate state means checkbox shows '-'
   other: boolean | 'indeterninate'
   meta: boolean
   structure: boolean
@@ -23,9 +25,8 @@ interface ComponentConfig {
 
 export const defaultConfig: ComponentConfig = {
   body: true,
-  // otherChecked has three states, 'true', 'false', 'indeterminate'
-  // indeterminate state means the checkbox shows a '-' instead of checked or unchecked
-  other: false as boolean | 'indeterminate',
+
+  other: false,
   meta: false,
   structure: false,
   readme: false
--- a/app/components/form/ComponentAndFormatPicker.tsx
+++ b/app/components/form/ComponentAndFormatPicker.tsx
@@ -1,0 +1,132 @@
+import React, { useState } from 'react'
+
+import { Dataset } from '../../models/dataset'
+
+import CheckboxInput from './CheckboxInput'
+import TristateCheckboxInput from './TristateCheckboxInput'
+import RadioInput from './RadioInput'
+
+interface ComponentAndFormatPickerProps {
+  dataset: Dataset
+  value: ComponentConfig
+  onChange: (v: ComponentConfig) => void
+}
+
+interface ComponentConfig {
+  body: boolean
+
+  other: boolean | 'indeterninate'
+  meta: boolean
+  structure: boolean
+  readme: boolean
+}
+
+export const defaultConfig: ComponentConfig = {
+  body: true,
+  // otherChecked has three states, 'true', 'false', 'indeterminate'
+  // indeterminate state means the checkbox shows a '-' instead of checked or unchecked
+  other: false as boolean | 'indeterminate',
+  meta: false,
+  structure: false,
+  readme: false
+}
+
+export const ComponentAndFormatPicker: React.FC<ComponentAndFormatPickerProps> = ({ dataset, value }) => {
+  const { meta, readme } = dataset
+  const enabledOtherComponents = ['structure'] // structure is always enabled
+  if (meta) enabledOtherComponents.push('meta')
+  if (readme) enabledOtherComponents.push('readme')
+
+  const [bodyFormat, setBodyFormat] = useState('csv')
+  const [config, setConfig] = useState(value)
+
+  const setConfigField = (field: string, value: boolean) => {
+    const update = { ...config, [field]: value }
+
+    if (field === 'other') {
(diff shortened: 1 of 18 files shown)


test/e2e/e2e.spec.ts
```diff
@@ -478,6 +478,18 @@ describe('Qri End to End tests', function spec () {
     await sendKeys('#dataset-name-input', "Enter")
   })
 
+  it('export CSV version', async () => {
```
uhLeeshUh, line 481: 🙌

app/components/form/ComponentAndFormatPicker.tsx
```diff
@@ -0,0 +1,132 @@
...
+interface ComponentAndFormatPickerProps {
+  dataset: Dataset
+  value: ComponentConfig
+  onChange: (v: ComponentConfig) => void
+}
+
+interface ComponentConfig {
+  body: boolean
+
+  other: boolean | 'indeterninate'
```
uhLeeshUh, line 18: small spelling error to `'indeterminate'`

app/components/form/ComponentAndFormatPicker.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+interface ComponentAndFormatPickerProps {
+  dataset: Dataset
+  value: ComponentConfig
+  onChange: (v: ComponentConfig) => void
+}
+
+interface ComponentConfig {
+  body: boolean
+
```
uhLeeshUh, line 17: any reason for this empty line?

app/components/form/ComponentAndFormatPicker.tsx
```diff
@@ -0,0 +1,132 @@
...
+  meta: boolean
+  structure: boolean
+  readme: boolean
+}
+
+export const defaultConfig: ComponentConfig = {
+  body: true,
+  // otherChecked has three states, 'true', 'false', 'indeterminate'
+  // indeterminate state means the checkbox shows a '-' instead of checked or unchecked
+  other: false as boolean | 'indeterminate',
```
uhLeeshUh, line 28: curious, why was the type assertion needed here?

app/components/form/ComponentAndFormatPicker.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+  return (
+    <div className='component-and-format-picker dialog-text-small'>
+      <div className='input-label'>Components</div>
+      <div className='options-row'>
+        <div className='options-column'>
+          <CheckboxInput
+            name='body'
+            checked={config.body}
+            onChange={() => { setConfigField('body', !config.body) }}
```
uhLeeshUh, line 75: small nit for these onChange props but probably don't need the inner set of curly braces, right?

app/components/modals/ExportDataset.tsx
```diff
@@ -1,91 +1,63 @@
...
+  const [exportMode, setExportMode] = useState('csv')
 
   const handleSubmit = () => {
     setDismissable(false)
     setLoading(true)
     error && setError('')
-    if (!onSubmit) return
-
-    alert('Exporting Files!')
+    exportDatasetVersion(version, exportMode)
```
uhLeeshUh, line 34: should this be wrapped in a try/catch so that if there's any error in sending the export event it can be captured here with `setError`? Or do we expect an error like that to be surfaced elsewhere in the app?

app/main.development.js
```diff
@@ -511,10 +510,17 @@ app.on('ready', () =>
...
+        let exportUrl = `${BACKEND_URL}/get/${refString}?format=zip`
+        if (config === 'csv') {
+          exportUrl = `${BACKEND_URL}/get/${refString}/body.csv`
+        }
         const win = BrowserWindow.getFocusedWindow()
-        await download(win, exportUrl, { filename, directory })
+        await download(win, exportUrl, {
+          filename,
+          directory,
+          openFolderWhenDone: true
```
uhLeeshUh, line 522: I like this touch of opening on completion

uhLeeshUh (approved):
LGTM! Just a few small comments

app/components/form/ComponentAndFormatPicker.tsx
b5: _kinda_. Adding the braces ensures the return value of on change is always `void`. That way if we mess with the return value of `setConfigField` or swap out the guts of the change handler, it won't "escape".

Pedantic? Yes. Intentional? Yes.

app/components/form/ComponentAndFormatPicker.tsx
b5: ugh I'm not a good typst

app/components/form/ComponentAndFormatPicker.tsx
b5: old habit from mutex locks in struct fields. The `other` field applies to all fields after it. Space implies `body` is not part of the `other` field. I'll add a comment

app/components/form/ComponentAndFormatPicker.tsx
b5: hrm, re-reading this, I don't think it's necessary. It was before I'd refactored this into it's own component. Will change

app/components/form/ComponentAndFormatPicker.tsx
uhLeeshUh: ah I see! Nice, good to know

app/components/modals/ExportDataset.tsx
b5: because it's wired to the inter-process-communication bus, export doesn't (easily) work with promise semantics. So, you're right that errors need to be handled elsewhere: specifically on the main electron process.

PR #644, opened 2020-10-08
feat(Collection): bulk remove datasets

### LINKED ISSUE #636: feat(bulk remove): wire up bulk remove action modal
- [x] allow for bulk removal of datasets
- [x] if an included dataset belongs to the sessionUser, show warning message
- [x] remove quick export CSV button

uhLeeshUh:
Closes #636


app/components/collection/collectionHome/DatasetList.tsx
```diff
@@ -101,36 +105,31 @@ export const DatasetListComponent: React.FC<DatasetListProps> = (props) => {
     ? `You have ${filteredDatasets.length} local dataset${filteredDatasets.length !== 1 ? 's' : ''}`
     : `Showing ${filteredDatasets.length} local dataset${filteredDatasets.length !== 1 ? 's' : ''}`
 
-  // use memoization to theoretically speed up rendering for example:
```
uhLeeshUh, line 104: @b5 I ended up removing this memoization logic since I noticed a bug with it where when a user selects a dataset => then types in filter text =>  then removes that filter text => now the original dataset has its checkbox selected but the bulk action buttons aren't present

b5: hrm, ok I guess we'll have to save it for another pass. The lag on building selections is not great, but I agree having things slow and working is better than having things fast and broken

app/components/modals/RemoveDataset.tsx
```diff
@@ -33,26 +29,42 @@ export const RemoveDatasetComponent: React.FunctionComponent<RemoveDatasetProps>
   const [error, setError] = React.useState('')
   const [loading, setLoading] = React.useState(false)
 
-  const handleChanges = (name: string, value: any) => {
+  const isLinked = !!datasets.find(dataset => dataset.fsiPath)
```
uhLeeshUh, line 32: I made the design decision here to show the checkbox to keep files if any of the selected datasets is linked to the file system. The action to keep files will be an all-or-nothing one for the full group of datasets, with only the ones with an `fsiPath` actually having any action taken to remove the files

b5: into it. good choice 😄

b5:
I think the answer to your question is to modify the signature of the function passed as the `OnSubmit` prop to the `RemoveDataset`, to make it more opaque, going from this:
```ts
export interface RemoveDatasetModal {
   // ..
   onSubmit: (...args: any) => any
}
```

to this:
```ts
export interface RemoveDatasetModal {
   // ..
   onSubmit: () => Promise<AnyAction>
}
```

you're already doing this wrapping inside the `handleSubmit` function of `RemoveDataset`, by moving it out into user land and requiring callers pass an opaque function that returns a promise, you can push configuration up to the call site, while retaining the requirement that the `onSubmit` prop  return a  promise so you can `await onSubmit()`

b5:
As for the principle problem you were having of a single checkbox bulk action, a set of 1 is still a set, and we should consider it a bulk action even though it's a single dataset. The fact that the same modal handles both singular and bulk (set-oriented) actions is a coincidence

uhLeeshUh:
@b5 along the lines of treating a set of 1 as still a set, do you have advice for differentiating how to correctly pass args to the `onSubmit` function [here](https://github.com/qri-io/desktop/pull/644/files#diff-c662b20f15225e0f651d2b2798629b85R46-R51)? The problem I've encountered is I don't have a reliable way of knowing if the single dataset to be removed was sent to the modal via the kabob menu or the bulk action button (this check of just "is there only one dataset" isn't sufficient). If via the kabob, I need to pass the args as a list to the `onSubmit` fn to match up with removeDatasetAndFetch, but if via the bulk action, then I treat it as the "set of one" approach and pass in an array to satisfy removeDatasetsAndFetch.

b5:
I'm saying build the `onSubmit` function before this handler is ever called, and pass it to `RemoveDataset`. You've got the problem figured out:

>  The problem I've encountered is I don't have a reliable way of knowing if the single dataset to be removed was sent to the modal via the kabob menu or the bulk action button

That's because this info is owned by the caller. So keep the `RemoveDataset` modal dumb and make it's `onSubmit` prop a function that the caller has to construct.

uhLeeshUh:
Ahhh gotcha. The only thing we'll need from the handler is the `keepFiles` value then but we can close over the other values in the correct format. Ok cool! I'll give that a go now

app/components/modals/RemoveDataset.tsx
```diff
@@ -60,24 +64,24 @@ export const RemoveDatasetComponent: React.FC<RemoveDatasetProps> = (props: Remo
       <div className='content-wrap'>
         <div className='content'>
           <div className='dialog-text-small'>
-            <p>Are you sure you want to remove <br/> <code>{username}/{name}</code>&nbsp;?</p>
+            <p>Are you sure you want to remove {datasetRefs}&nbsp;?</p>
             <br/><br/>
-            {sessionUsername === username && <div className='warning'>Warning: removing a dataset which belongs to you means you cannot return to that dataset&apos;s history.</div>}
+            {isDeletingSessionUserDataset && <div className='warning'>Warning: removing a dataset which belongs to you means you cannot return to that dataset&apos;s history.</div>}
```
ramfox, line 69: how do we feel about changing this text from

"a dataset which belongs to you"
to
"a dataset that you created"

app/components/collection/collectionHome/TableRowHamburger.tsx
```diff
@@ -1,18 +1,23 @@
...
 import { VersionInfo } from '../../../models/store'
 import { Modal, ModalType } from '../../../models/modals'
+import { removeDatasetAndFetch } from '../../../actions/api'
 import { setModal } from '../../../actions/ui'
 import { connectComponentToProps } from '../../../utils/connectComponentToProps'
 
 interface TableRowHamburgerProps {
   data: VersionInfo
   setModal: (modal: Modal) => void
+  removeDatasetAndFetch: (...args: Parameters<typeof removeDatasetAndFetch>) => Promise<AnyAction>
```
ramfox, line 14: I like this syntax! 

On one hand, I don't like that you can't just glance at the props to understand what the function is expecting, but on the other hand it means we are ensuring the function has the parameters we are expecting & its one fewer place to change if something about the action changes.

Plus, we all have IDE's that will be able to give us quick looks into what parameter's the action is expecting.

app/components/collection/collectionHome/DatasetList.tsx
```diff
@@ -101,36 +105,41 @@ export const DatasetListComponent: React.FC<DatasetListProps> = (props) => {
...
+    const refs = selected.map(qriRefFromVersionInfo)
+    return handleBulkActionForDatasets('pull', 'pulling', 'pulled', pullDatasets, refs)
+  }
+
+  const handleBulkRemoveSelectedDatasets = async (keepFiles: boolean) => {
+    const refs = selected.map(ds => ({ username: ds.username, name: ds.name, isLinked: !!ds.fsiPath, keepFiles }))
+    return handleBulkActionForDatasets('remove', 'removing', 'removed', removeDatasetsAndFetch, refs)
+  }
+
+  const handleBulkActionForDatasets = async (actionType: DatasetActionType, actionGerund: string, actionPastTense: string, action: DatasetAction, refs: Array<{username: string, name: string, isLinked?: boolean, keepFiles?: boolean}>) => {
```
ramfox, line 118: we aren't using `keepFiles` here, I think we should remove it!

app/components/collection/collectionHome/DatasetList.tsx
```diff
@@ -101,36 +105,41 @@ export const DatasetListComponent: React.FC<DatasetListProps> = (props) => {
...
-        <button disabled={bulkActionExecuting} onClick={() => alert(`quick export CSV: ${selected}`)}>Quick Export CSV</button>
-        <button disabled={bulkActionExecuting} onClick={() => alert(`remove: ${selected}`)}>Remove</button>
-      </>}
-    </div>)
-  }, [selected, filteredDatasets, bulkActionExecuting])
+  const openRemoveModal = () => {
+    setModal(
+      {
+        type: ModalType.RemoveDataset,
+        datasets: selected.map(ds => ({ username: ds.username, name: ds.name, fsiPath: ds.fsiPath })),
```
ramfox, line 138: I think we should be moving around `VersionInfo`s rather than a new datatype! Let's just pass the list of selected VersionInfos, and then push the `selected`

And if we are trying to use a generalized `handleBulkActionForDataset` it should probably take `VersionInfo`s instead of `QriRefs` because qriRefs will only give us the barest of information. This means that we would also have to adjust the `pullDatasets` API function to take `VersionInfo[]` as well, & adjust the definition of a `DatasetAction` in this file.

app/components/collection/collectionHome/DatasetList.tsx
ramfox: I write somewhere below that I think we should be passing around `VersionInfo`s, not new types here, if so, we should change the `Array<{...}>` to `VersionInfo[]`

app/components/collection/collectionHome/DatasetList.tsx
```diff
@@ -101,36 +105,41 @@ export const DatasetListComponent: React.FC<DatasetListProps> = (props) => {
...
-          setBulkActionExecuting(false)
-          openToast('error', 'pull-error', `pulling datasets: ${reason}`)
-        })
+  const handleBulkPullSelectedDatasets = async () => {
+    const refs = selected.map(qriRefFromVersionInfo)
+    return handleBulkActionForDatasets('pull', 'pulling', 'pulled', pullDatasets, refs)
+  }
+
+  const handleBulkRemoveSelectedDatasets = async (keepFiles: boolean) => {
+    const refs = selected.map(ds => ({ username: ds.username, name: ds.name, isLinked: !!ds.fsiPath, keepFiles }))
```
ramfox, line 114: as stated below (or whatever order these comments appear in) I think we should be passing around `VersionInfos` rather than new types, just pass `selected` into refs.

app/components/collection/collectionHome/DatasetList.tsx
```diff
@@ -101,36 +105,41 @@ export const DatasetListComponent: React.FC<DatasetListProps> = (props) => {
...
-        .then(() => {
-          setBulkActionExecuting(false)
-          openToast('success', 'pull-success', `pulled ${refs.length} ${refs.length === 1 ? 'dataset' : 'datasets'}`)
-        })
-        .catch((reason) => {
-          setBulkActionExecuting(false)
-          openToast('error', 'pull-error', `pulling datasets: ${reason}`)
-        })
+  const handleBulkPullSelectedDatasets = async () => {
+    const refs = selected.map(qriRefFromVersionInfo)
```
ramfox, line 109: same comment as below, let's just use `VersionInfo` for all these bulk functions.

app/components/collection/collectionHome/DatasetList.tsx
ramfox: also it looks like `isLinked` is getting calculated in the `RemoveModal` so we don't have to worry about losing any data by turning this into a `VerisonInfo`

app/models/modals.ts
```diff
@@ -45,9 +47,8 @@ export interface PublishDatasetModal {
 
 export interface RemoveDatasetModal {
   type: ModalType.RemoveDataset
-  username: string
-  name: string
-  fsiPath?: string
+  datasets: Array<{ username: string, name: string, fsiPath?: string }>
```
ramfox, line 50: Think this should be `VersionInfo[]`

ramfox (changes requested):
Looks great! Love love love the tests & the util function. It's definitely made me realize some stuff we can refactor in the e2e tests :)

I have some strong opinions about what types we should be using for the bulk actions, especially since we are trying to generalize some of the work so it can be used for multiple kinds of bulk actions! Basically, I think we should be using `VersionInfo[]`s everywhere, since that is the data type we start with and is the most general we will ever need to be for any actions we need to do.

ramfox:
Another note, I don't think your tests will pass until the `rename` pr is merged, I had to do a lot of finessing on tests that relied on elements that had changed! But the tests that you added definitely work.

app/components/collection/collectionHome/TableRowHamburger.tsx
uhLeeshUh: sweet! Yeah I'm a big fan of it as well because it makes keep track of changed types a lot easier

uhLeeshUh:
Oh yeah @ramfox nice idea to simplify the passed action type to `VersionInfo` -- I was finding it hard to keep track of those two similar types and think `VersionInfo` as a used pattern for all our actions is great

app/components/collection/collectionHome/DatasetList.tsx
uhLeeshUh: Hmm my only worry with taking away `isLinked` here is that the calculation in `RemoveDataset` is pretty basic (`isLinked` is true if any dataset in the set has an `fsiPath`) and I worry that in each individual `removeDataset` call that means we'd try to pass`keep-files` where it's not actually needed and it could error, right?

app/components/collection/collectionHome/DatasetList.tsx
```diff
@@ -101,36 +104,41 @@ export const DatasetListComponent: React.FC<DatasetListProps> = (props) => {
...
-          setBulkActionExecuting(false)
-          openToast('error', 'pull-error', `pulling datasets: ${reason}`)
-        })
+  const handleBulkPull = async () => {
+    const actionCallback = async () => pullDatasets(selected)
+    return handleBulkActionForDatasets('pull', 'pulling', 'pulled', actionCallback, selected)
+  }
+
+  const handleBulkRemove = async (keepFiles: boolean) => {
+    const actionCallback = async () => removeDatasetsAndFetch(selected, keepFiles)
```
uhLeeshUh, line 113: @ramfox what do you think of this `actionCallback` pattern? I figured it was a way for us to still reuse the `handleBulkActionForDatasets` function but give both the `handleBulkRemove` and `handleBulkPull` handlers the ability to craft their own action callers (and in the case of remove, add `keepFiles` as the final arg)

ramfox: I think this is great!

ramfox (approved):
LGTM!!!

Nice :)

PR #645, opened 2020-10-08
feat(RenameDataset): add `RenameDataset` modal and rename dataset button to dataset page

### LINKED ISSUE #637: Dataset Rename Modal
This issue will add `Rename Dataset` as an action, available in the hamburger menu on each row in the collection view and in the header of the dataset view.

As the name suggests, the user will see a modal with a text input containing the current dataset name, where they can edit and click "Rename".  The modal should validate the input and not allow the user to rename to an invalid name.  Upon success/failure, a toast should appear.

Other items:
- [ ] update e2e tests for rename
- [x] remove the existing dataset name display/rename UI that currently appears at the top of the history list.

ramfox:
closes #637 

Still todo: e2e test fixes

plz read commits for thought process!

### DIFF
--- a/app/components/collection/layouts/DatasetLayout.tsx
+++ b/app/components/collection/layouts/DatasetLayout.tsx
@@ -2,12 +2,14 @@ import React from 'react'
 import { faDownload, faTrash } from '@fortawesome/free-solid-svg-icons'
 import { RouterProps } from 'react-router'
 
-import Store from '../../../models/store'
+import Store, { VersionInfo } from '../../../models/store'
 import { connectComponentToPropsWithRouter } from '../../../utils/connectComponentToProps'
 import { setModal, setSidebarWidth } from '../../../actions/ui'
 import {
+  selectFsiPath,
   selectSidebarWidth
 } from '../../../selections'
+import { removeDatasetsAndFetch } from '../../../actions/api'
 import { Modal, ModalType } from '../../../models/modals'
 import { QriRef, qriRefFromRoute } from '../../../models/qriRef'
 
@@ -16,11 +18,13 @@ import Layout from '../../Layout'
 import LinkButton from '../headerButtons/LinkButton'
 import PublishButton from '../headerButtons/PublishButton'
 import RenameButton from '../headerButtons/RenameButton'
+import { ApiAction } from '../../../store/api'
 
 interface DatasetLayoutProps extends RouterProps {
   // from connect
   sidebarWidth?: number
   qriRef?: QriRef
+  versionInfo?: VersionInfo
   onSidebarResize?: (width: number) => void
   setModal: (modal: Modal) => void
   // from props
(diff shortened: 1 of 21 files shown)


app/components/collection/layouts/DatasetLayout.tsx
```diff
@@ -43,13 +45,21 @@ const DatasetLayoutComponent: React.FunctionComponent<DatasetLayoutProps> = (pro
...
-    { type: 'component', component: <PublishButton key='publish-button' /> }
+    { type: 'component', component: <PublishButton key='publish-button' /> },
+    { type: 'component', component: <RenameButton key='rename-button' /> },
+    {
+      type: 'button',
+      id: 'remove-dataset',
+      icon: faTrash,
+      label: 'Remove',
+      onClick: () => { setModal({ type: ModalType.RemoveDataset, username: qriRef.username, name: qriRef.name }) }
+    }
```
ramfox, line 62: I'm aware that it's odd that I added a RenameButton component for one task and an inline button creation for remove dataset, but my next pr is going to be to get much of these in a `hamburger` & I wanted to test out both options.

b5: this was my immediate question. thank you.

chriswhong (changes requested):
This modal's input is styled differently from other modals and I think we should be consistent here.  The full red background on invalid input is also kind of biting, and suggest we make the outline of the input red and show the message in red.


Another suggestion:
- Instead of an action button where the other buttons are, just add a pencil icon to the right of the dataset name.  Rename isn't quite the same level of action as an export or push

We should also add rename to the hamburger menu in collection view.  I think adding a pencil icon to the dataset name would work here as well (versus the hamburger menu where the other actions live)


Qri_Desktop

b5:
> Instead of an action button where the other buttons are, just add a pencil icon to the right of the dataset name. Rename isn't quite the same level of action as an export or push

While I agree this might be nice, I'm not into the idea of adding a new UI convention until we've had a chance to think it through. To me the pencil icon implies that I'd be able to edit the dataset name _without_ a modal – editing the title directly in the `NavTopbar` component, which we've just done a bunch of work to move into a modal.

Renaming a dataset isn't–and shouldn't be–a common task. Let's push it into the hamburger menu & make the modal styling changes you've suggested

app/components/collection/layouts/DatasetLayout.tsx
```diff
@@ -49,7 +59,21 @@ const DatasetLayoutComponent: React.FunctionComponent<DatasetLayoutProps> = (pro
...
+      icon: faTrash,
+      label: 'Remove',
+      onClick: () => {
+        setModal({
+          type: ModalType.RemoveDataset,
+          datasets: [{ ...qriRef, fsiPath }],
+          onSubmit: async (keepfiles: boolean) => removeDatasetsAndFetch([{ ...qriRef, fsiPath }], keepfiles)
+        })
+      }
+    }
```
ramfox, line 76: I can't express how much I hate this, hahaha.

So when I refactor the datasetlayout next I'm going to move this into its own button.

@b5  don't hate me! But I think it's this weird logic/state tree mush that we can push off into a button that should be able to be placed wherever, & won't clog up any parent components as this one is currently doing here.

chriswhong (approved):
🚀

Issue #647, opened 2020-10-09
Selected dataset count gets out of sync when removing datasets from Collection view


uhLeeshUh:
### What version of qri are you using (`qri version`)? 
v0.9.13-dev

### What version of Qri Desktop are you using?
0.4.6-dev

### What did you do?
- Add 3 local datasets
- Navigate to Collection view
- Select 2 of the datasets via the checkbox
- Click "Remove" button and select "Remove" on the modal
- UI _appears_ to have worked successfully
- Now select the final remaining dataset via checkbox and the copy reads "3 selected"

(note, the same issue happens without bulk remove. If you select a dataset before removing it via the hamburger menu, it persists in table selection state)

### What happened?
The internal `selected` state of the Collection table jumped from the correct number of selected datasets to, after selecting another dataset, the incorrect number (which includes the deleted datasets)

### What did you expect to happen?
That those deleted datasets would not be considered selected anymore

### Do you have a suggested fix?

The problem is likely that there is a discrepancy between internal component state of the `DatasetList` component and the `ReactDataTable` component. This can be visible by logging the value of `selectedRows` in `handleSelectedRowsChange` of the `DatasetList` component

PR #648, opened 2020-10-09
fix(DatasetList): fix of Collection selected table state

uhLeeshUh:
Addresses #647

### DIFF
--- a/app/components/collection/collectionHome/DatasetList.tsx
+++ b/app/components/collection/collectionHome/DatasetList.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useCallback } from 'react'
+import React, { useState, useCallback, useEffect } from 'react'
 import { Action, AnyAction } from 'redux'
 import classNames from "classnames"
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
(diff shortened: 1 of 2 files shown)


app/components/collection/collectionHome/DatasetList.tsx
```diff
@@ -44,6 +44,20 @@ export const DatasetListComponent: React.FC<DatasetListProps> = (props) => {
   const [onlySessionUserDatasets, setOnlySessionUserDatasets] = useState(false)
   const [bulkActionExecuting, setBulkActionExecuting] = useState(false)
 
+  useEffect(() => {
+    if (datasets.length === 0) return
```
uhLeeshUh, line 48: There is a beat in the render cycle where datasets clear before coming back with the correct amount (orig - removed)

app/components/collection/collectionHome/DatasetList.tsx
```diff
@@ -56,6 +70,7 @@ export const DatasetListComponent: React.FC<DatasetListProps> = (props) => {
 
   // keep track of selected items in state
   const handleSelectedRowsChange = useCallback(({ selectedRows }: { selectedRows: VersionInfo[] }) => {
+    console.log({ selectedRows }) // THIS VALUE IS INCORRECT
```
uhLeeshUh, line 73: This is where the problem is manifesting, but I'm not sure how to interact with the internal state of ReactDataTable to get it to reflect the `selected` value here

ramfox (approved):
@uhLeeshUh great explanation of the issue! I was able to look into the `ReactDataTable` docs and figure out a fix!

Here is what I looked at if you are curious: https://www.npmjs.com/package/react-data-table-component#selectable-rows

PR #651, opened 2020-10-09
fix(NavbarButtons): ensure correct buttons show & are in the hamburger

ramfox:
Making small button components for each button w/ separate logic. Each manages its own logic & will appear when appropriate & can be moved around and added without having to rewire a parent component.

My largest motivation for doing this is we've learned as the codebase has grown and changed that the best way to minimize buggy refactors and minimize the time it takes to do refactors is to minimize duplicate code & try to make any component with complicated logic be responsible for itself. By having a button for each action, the buttons themselves can be responsible for ensure they appear at the right time, and not having to have a parent component that figures out the complicated logic for showing or hiding each button. And again, as I mentioned above, this also means that if the buttons need to move around we don't have to do large refactors to move them.


app/actions/api.ts
```diff
@@ -641,52 +641,32 @@ export function removeDataset (
...
     } catch (action) {
       if (!action.payload.err.message.contains('directory not empty')) {
         throw action
       }
-      dispatch(openToast('error', 'removeDatasetsAndFetch', action.payload.err.message))
+      dispatch(openToast('error', 'remove-error', `error removing ${datasetsString}: ${action.payload.err.message}`))
+      dispatch(setBulkActionExecuting(false))
       return Promise.reject(action.payload.err.message)
     }
   }
```
ramfox, line 672: I'd love a future refactor that makes reading and writing actions a bit easier to decipher, maybe some helper that takes optional expected parts of the action and wires it up for you & handles errors.

dustmop (approved):
LGTM

I'm not an expert on desktop's codebase, and there's a lot of code moving around here, but after reading over it it looks good. I especially appreciate the description in https://github.com/qri-io/desktop/pull/651/commits/bb36b49fd956d48932087f01ace97b96735c7fc2 that explains the important refactoring happening there.

PR #653, opened 2020-10-12
fix: remove fixed height on input descriptions

chriswhong:
Removes fixed height for input description text (prevents scrollbar in modals)


dustmop (approved):
LGTM

PR #654, opened 2020-10-12
fix(DatasetList): ensure overlay menu can be seen for last item

ramfox:


Arqu (approved):
Version is locked, so LGTM.

PR #668, opened 2020-11-03
feat(remote): add build flag and special logic for a remote webapp

### LINKED ISSUE #663: feat(*): compile remote app as a webapp w/o electron
We want to be able to use qri desktop without electron! 

* [x]  parse out calls to electron & place in separate files. Add equivalent actions for if the app is loaded via the web.
* [x] add webpack files to compile the app
* [x] add `remote` flag that doesn't load the network, editing, or login
* [x] build webapp & replace the currently hosted app

ramfox:
Only shows the collection page, the dataset page, and the export modal.

Known bugs:
-  FIXED - need to figure out a way to have the dev-mode remote point to the local api & the prod-mode remote point to the browser origin API. Currently, if you want to test this locally, you need to make sure that your `app/store/api.ts` file has the `url` var point to the `BACKEND_URL/endpoint`, and inside the `ExportModal`, inside the `exportLink` func, the `exportUrl` should point to the `BACKEND_URL` const as well.

- filled [issue](https://github.com/qri-io/desktop/issues/669) & punted - export modal needs to dismiss after download

closes #663


app/components/collection/headerButtons/ExportButton.tsx
```diff
@@ -54,8 +54,8 @@ export default connectComponentToPropsWithRouter(
   (state: any, ownProps: ExportButtonProps) => {
     const qriRef = qriRefFromRoute(ownProps)
     return {
-      ...ownProps,
-      qriRef
+      qriRef,
+      ...ownProps
```
ramfox, line 58: need to ensure that if we pass in our own props, it doesn't get overridden with props from state. This is specifically important when we are looking at export, since we could be exporting from the collection page, in which case it is not possible to derive `qriRef` props from the route.

b5 (approved):
Very nice. Let's get this merged.

The number of times `__BUILD__` has to show up to support two compilation targets adds a bunch of complexity that I'd love to tame over time. Short term answers:
* rely on electron-specific APIs as little as possible (duh, but we now have a metric we can measure)
* write a custom CI-check that ensures any import that uses `TARGET_PLATFORM` has a uniform implementation in both platforms

But all of that's future work. Let's merge!

PR #671, opened 2020-11-10
fix(backend): kill any `qri connect` processes that aren't compatible

ramfox:
new dependancy `fkill`: https://www.npmjs.com/package/fkill

In this commit, we refactor our `checkNoActiveBackendProcess` to also kill any processes at port `2503` if we aren't able to figure out the version number or if the version number it returns is not the expected version.

### DIFF
--- a/app/backend.js
+++ b/app/backend.js
@@ -6,8 +6,13 @@ const os = require('os')
 const http = require('http')
 const fkill = require('fkill')
 const { dialog } = require('electron')
+const findProcess = require('find-process')
 
 const { backendVersion: expectedBackendVer } = require('../version')
+
+const { PORT } = require('./constants')
+
+const errPortOccupied = new Error('port-occupied')
 
 // BackendProcess runs the qri backend binary in connected'ed mode, to handle api requests.
 class BackendProcess {
@@ -115,9 +120,41 @@ class BackendProcess {
   async checkNoActiveBackendProcess () {
     log.info("checking for active backend process")
 
+    let errorCheckingPort = false
+    const list = await findProcess('port', PORT)
+                        .catch(e => {
+                          log.info(`error examining port ${PORT}: ${e}. We will attempt to ping the endpoint to see if we get a response`)
+                          errorCheckingPort = true
+                        })
+
+    if (!errorCheckingPort) {
+      if (list.length === 0) {
+        // port is clear
+        return
+      }
+
+      list.forEach(process => {
+        if (!process.cmd.includes('qri connect')) {
+          /**
+           * if the process at the port is not a qri process, throw an error
+           */
+          throw errPortOccupied
+        }
+      })
+    }
+
     const healthResponse = async () => {
       return new Promise((resolve) => {
-        http.get('http://localhost:2503/health', (res) => {
+        http.get(`http://localhost:${PORT}/health`, (res) => {
+          if ( errorCheckingPort && res.statusCode !== 200 ) {
+            /**
+             * if we do not have a successful status code AND we were not able
+             * to confirm that a qri process is running at the given port, we should
+             * warn the user & have them close the process on their own, or reach
+             * out to us for help.
+             */
+            throw errPortOccupied
+          }
           let body = ''
           res.on('data', (chunk) => {
             body += chunk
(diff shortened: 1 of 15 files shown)


b5:
the sum result of these changes is a program that's hostile to _any_ process running on port 2503, which is bad. We don't know what's running on 2053, might not the Qri binary.

We need to do two additional pieces of work:
* add function that determines desktop's configured qri repo path `repoPath(): string`. Right now that should be a deterministic function that  when running in production mode returns `$HOME/.qri`(we'll add configuration later).
* write a function that reads the configured API address from `path.join(repoPath(), 'config.yaml')`, and use that address to pick up  the configured port number, then feed _that_ number to the `fkill` function you've written here

Irritating? Yes? But a really important step to avoid some potentially terrible side effects

ramfox:
Yes but the entire app is hardcoded to be looking at port `2503`, so if the process running at that port isn't the qri binary, the Desktop can't function anyway. Also, just because the qri config says that it's running at a different port, doesn't change the fact that we could be killing something that isn't the qri binary.

Did a little bit of research and found an npm package [find-process](https://www.npmjs.com/package/find-process) that will give you info about a process running at a particular port! I didn't realize this was an option or I would have used it yesterday.

My proposal now is:
1) check if a process is running at port `2503` (for now. When desktop is more configurable that value should be the configured port, not hardcoded)
2) if nothing is running at that port, we are all good & can return
3) check if the process that is being run IS `qri connect`
4) if it's not, throw an error, that will be caught in the main process. The main process will then emit an event ('port occupied' w/ the port #) & the renderer process (aka App.tsx) will be listening for that error, and will display an error message to the user. 
"Qri Desktop cannot start because another process is occupying port `2503`. Please close that process or reach out to us on Discord for help!" 
5) if it is a `qri connect` process, keep the healthCheck/`fkill` logic already in this pr

If I can get a sanity check on this, I'll code it up!

b5:
even better. I still think we need to get past the hardcoded `2503` thing, but confirming the process operating on  2503 is in fact called `qri connect` solves the main problem. Now we just need to make sure `find-process` works on windows 😄. If so, let's  do it & skip the configuration checking for another day!

ramfox:
It claims it works on mac, windows, linux, and android (???)

Based on issues filed, looks like it works fine on windows, but not for Linux machines that don't have `netstat` installed

app/backend.js
```diff
@@ -109,20 +119,89 @@ class BackendProcess {
...
     log.info("checking for active backend process")
-    const healthCheck = async () => {
-      return new Promise((res, rej) => {
-        http.get('http://localhost:2503/health', (data) => {
-          res(true)
+
+    let errorCheckingPort = false
+    const list = await findProcess('port', PORT)
+                        .catch(e => {
+                          log.info(`error examining port ${PORT}: ${e}. We will attempt to ping the endpoint to see if we get a response`)
```
b5, line 126: nit: `we will` implies there's a group of... somethings inside my computer. Prefer active-voice,  third-person:

```
error examining port ${PORT}: ${e}. pinging health endpoint for a response...
```

app/backend.js
```diff
@@ -109,20 +119,89 @@ class BackendProcess {
...
+                          errorCheckingPort = true
+                        })
+
+    if (!errorCheckingPort) {
+      if (list.length === 0) {
+        // port is clear
```
b5, line 132: if you use an ellipsis in the `log.info` call above add `log.info('success')` here to indicate a healthy ping

app/components/PortOccupied.tsx
```diff
@@ -0,0 +1,20 @@
...
+        <ExternalLink
+          id='discord'
+          href={DISCORD_URL}
+        > Discord </ExternalLink>
+      if you have any questions!</div>
+    </div>
+  </WelcomeTemplate>
+)
+
+export default PortOccupied
```
b5, line 20: This is lovely

app/constants.js
```diff
@@ -1,6 +1,7 @@
 // these constants may be used by both the main process and frontend
 // so they need to be defined as CommonJS
 
+const PORT = 2503
```
b5, line 4: `PORT` can mean a lot of things. how about `BACKEND_PORT`?

app/main.development.js
```diff
@@ -534,6 +535,10 @@ app.on('ready', () =>
         quitting = true
       })
 
+      app.on('render-process-gone', () => {
+        app.quit()
+      })
```
b5, line 540: 🔥

b5 (approved):
CI has two linting beefs:

```
yarn run v1.22.4
$ ./node_modules/.bin/eslint --ext tsx --ext ts app/

/home/circleci/desktop/app/components/collection/collectionHome/DatasetsTable.tsx
  131:13  error  Component definition is missing display name  react/display-name

/home/circleci/desktop/app/components/modals/ExportSubmitButton.tsx
  1:8  error  'React' is defined but never used  @typescript-eslint/no-unused-vars

✖ 2 problems (2 errors, 0 warnings)
```


Other than that, this is _awesome_. Pumped to have a much better solution in place

app/backend.js
ramfox: hahahAHAHAH

PR #674, opened 2020-12-01
feat(change report flow): add `View Changes` button & change report route

ramfox:
- adds `/collection/changereport` route. For now, this directs to a `TempChangeReport` page
- adds `ViewChangesButton` - this pushes the correct versions to the router
- adds `ViewChangesButton` to storybook
- adds the Change Report to the `save` flow. If you are saving a dataset that has a previous version, it will direct you to the 'change report' rather than the dataset page

Made changes to the e2e tests to compensate for the save flow changes, however, it looks like the e2e tests are currently broken. Even at master. Those fixes should happen in a different pr, though.


app/models/qriRef.ts
```diff
@@ -77,6 +77,28 @@ export function refStringFromQriRef (qriRef: QriRef): string {
   return route
 }
 
+// qriRefFromString takes a string and turns it into a qriRef
+// based on two possible formats for the ref strings:
+// [username]/[name]/at/[network]/[path]
+// [username]/[name]
+// if the string does not follow either strict formatting, we return `undefined`
+export function qriRefFromString (refString: string): QriRef | undefined {
```
ramfox, line 85: added this when I thought we would need to parse the ref strings. The way the flow is set up right now, we don't need to, but leaving this in case it is useful in the future.

app/components/collection/headerButtons/ViewChangesButton.tsx
```diff
@@ -0,0 +1,96 @@
...
+  const right = qriRef
+  let left
+  if (qriRef.path === '') {
+    left = log[0]
+  } else {
+    const rightIndex = log.findIndex(vi => vi.path === qriRef.path)
+    // since we have already guarded against the given path being the last index
+    // we can add 1 to the rightIndex without worrying about overflow
+    left = log[rightIndex + 1]
+  }
```
b5, line 65: This is logic that needs to be abstracted away from the component into a selector that returns either nothing, or a data structure that can be fed directly to this component. The closest thing we have to the data structure we want is [`DiffParams`](https://github.com/qri-io/qri/blob/b683443616cd3f0138b6096684456e6d44727e38/lib/diff.go#L24-L39
) on the go side. Thinking in pseudocode for a second here:

```typescript 
// define a data structure that represents configuration for a change report request
interface ChangeReportParams {
  left: qriRef
  right: qriRef
}

// ... in a selectors file...

// a selector that produces that configuration from state
// returns undefined if no report is possible
function selectChangeReportParams(state): ChangeReportParams|undefined {
  // most of the highlighted code from this component.
  // you'll need to adjust params for this function to match
}

// ... in this component file ...

// adjust props to be the "data" param + visual configuration
interface ViewChangesProps extends RouteProps {
   data?: ChangeReportParams
   showIcon: boolean
   size: 'sm' | 'md'
 }

// return null if data is undefined. Later we could show "greyed out" UI when data prop is undefined
export const ViewChangesButtonComponent: React.FunctionComponent<ViewChangesProps> 
```

b5 (commented):
I think this could be re-worked to be a little more testable / workable. Bones look great

app/components/collection/headerButtons/ViewChangesButton.tsx
b5: I should also point out this `ChangeReportParams`  would be the same data structure spit out by any UI that allows compare-view configuration

.eslintrc
```diff
@@ -25,6 +25,7 @@
     "@typescript-eslint/prefer-readonly": "off",
     "@typescript-eslint/no-var-requires": "off",
     "@typescript-eslint/quotes": "off",
+    "@typescript-eslint/consistent-type-assertions": "off",
```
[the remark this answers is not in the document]
ramfox: Allows us to use the syntax `{} as SomeInterfaceOrType` for type assertion, which I personally find the clearest type assertion syntax. Not married to it, tho, so can change it.

app/components/collection/headerButtons/ViewChangesButton.tsx
```diff
@@ -0,0 +1,96 @@
...
+  const icon = <span className="fa-layers fa-fw">
+    <FontAwesomeIcon icon={faSquare} transform='grow-10' />
+    <FontAwesomeIcon color='white' icon={faPlus} transform='up-3 shrink-2' />
+    <FontAwesomeIcon color='white' icon={faMinus} transform='down-6 shrink-2'/>
+  </span>
+
+  return (<HeaderColumnButton
+    id='view-changes'
+    label='View Changes'
+    tooltip='Explore the changes made from the previous version of this dataset to the version you are currently viewing'
```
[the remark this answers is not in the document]
ramfox: thank you!

.eslintrc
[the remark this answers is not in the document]
b5: ah ok, then I think we should follow the advice on this stack overflow post instead:

https://stackoverflow.com/questions/49741673/typescript-initiate-an-empty-interface-object/49741799

`{} as Foo` will lead to issues down the road. The assertion we're making here creates an edge condition the type-checker can't catch, because `{}` isn't actually a `Foo` if foo has any specified fields.

Instead, let's define constructors in the same files types are defined.

app/components/collection/headerButtons/ViewChangesButton.tsx
ramfox: This is great. Makes it harder to test the logic, because we can't use storybook to verify (and our state tree/selectors testing is subpar), but it's a much cleaner experience to read the code.

Slating this kind of refactor for the other buttons that have specific logic.

b5: we can (eventually) test those bits of logic with unit tests, which are much cheaper to run than storybook. Delighted to hear you're into the change

b5 (approved):
LGTM!

PR #675, opened 2020-12-01
refactor(`Changes`): add actions, reducer, selectors, & `Changes` Component

ramfox:
Connect to the backend to fetch the Change Report!

This pr contains all the connective tissue to get the change report data into the desktop & display it.

This also adjusts the way we get handle `qri version`. The backend has added more information to its response and we need to adjust how we parse it.


b5 (approved):
rubber stamp!

PR #676, opened 2020-12-02
refactor(`Stats`): add stats as a first class component!

ramfox:
- adjusts the selectors, reducers, and actions
- adjust the `Body`, `BodyTable`, & `DetailBar`components to work properly, even if there are no stats available
- use the `StatDiffItem` rather than the `StatsChart` component inside the `DetailsBar`


b5 (approved):
rubber stamp!

PR #677, opened 2020-12-04
refactor(`DatasetChanges`): adjust layout and routes for `Changes`

ramfox:
Adjusted the layout and routes for the `Changes` page so that it shows up inside our typical dataset layout.

I'm sure we are going to adjust this further in the future!

This also adjusts our interfaces/expectations for the `/changes/` response to include `version_info`, which we map to `versionInfo` on intake + adding the `commitTitle` from the `commit` component


b5 (approved):
Looks good!

PR #679, opened 2020-12-10
fix(`Changes`): first round styling adjustments based on feedback

ramfox:
- StatusDot - when 'unmodified' we now have an option to show text that indicates that there are "No Changes"
- Segment - optional `startOpen` param that defaults to `true`. When `false`, the `Segment` starts closed
- StringDiff - when a component is "unmodified", default to having all text displayed
- ChangeReport - “Base” is red “Compare” is green
- StringDiff - Structure should by default start "rolled up" aka `startOpen` === false
- Labeled stats - Now we expect `Stat` to have another field `inBytes`. When a `Stat` has a `number` value & `inBytes` is true, we display the stat in Gb, Mb, b, etc.
- SummaryItem - when we iterate through the given Stats, if one of the stats is "totalSize", send over "File Size" as the label, and ensure `inBytes` is `true`.
- DatasetChanges - adjust layout so changes feels more like a modal that you need to click back from
- DatasetSummaryItem - add optional `shouldDisplayName` param, when false does not show the `username/dataset` name combo

### DIFF
--- a/app/components/Layout.tsx
+++ b/app/components/Layout.tsx
@@ -1,20 +1,25 @@
 // a layout component for the resizable sidebar with main content area
 import React from 'react'
 import { Resizable } from './Resizable'
-import NavTopbar from '../components/nav/NavTopbar'
+import NavTopbar, { NavbarButtonProps } from '../components/nav/NavTopbar'
 
 interface LayoutProps {
   id: string
-  sidebarWidth: number
+  sidebarWidth?: number
   sidebarContent?: React.ReactElement
   mainContent: React.ReactElement
   headerContent?: React.ReactElement
-  topbarButtons: NavbarButtonProps[]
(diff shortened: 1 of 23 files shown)


chriswhong (approved):
🏄‍♂️ 

Each bullet point in your PR notice works as expected.  Some observations:

* Horizontal resize is *really* choppy, I am sure this is due to many charts being rendered and needing to calculate their full-widthiness
* Still seeing negative values for charts for numeric columns
* Overlapping labels on skinny histogram bars - can we exclude them if we know they will be below a certain width?

ramfox:
for resize -> we can also just turn off `animated` resizing for the `change report` page
negative values -> if this is on a chart, then that's a backend issue unfortunately
labels -> we can add logic to exclude labels. Do you mean we shouldn't show labels in cases where the labels are too long or the bars in the chart are too skinny?

